PR #25, opened 2022-09-08
Fix Rescue Prime

aszepieniec:
This pull request changes the way hashing works in twenty-first. First, it introduces a reference implementation in SageMath for Rescue-Prime, and a compliant rust implementation called `RescuePrimeRegular`. Second, it refactors the `Hasher` trait to include a new data type field `T` which represents the atomic unit of the hash function's input and output vectors. Third, it modifies existing tools to use the new trait and type. Fourth, it depracates the old `RescuePrime` module in favor of `RescuePrimeRegular`.

There are some remaining tasks that belong in this pull request. However, many elements are already in place. In particular, the refactor of the `Hasher` trait is ready for review.

Remaining tasks:
 - [x] drop `RescuePrimeDeprecated` and fix references
 - [x] drop `RescuePrimeXlix` and fix references
 - [x] revise `MPolynomial::mod_pow`
 - [x] revise `hash_with_salts`


Sword-Smith:
We might be able to get rid of `hash_with_salts`. It was used for salted Merkle trees, but we might be able to provide the salts another way or let the caller handle the salts such that the salted Merkle trees can be scrapped alltogether.

twenty-first/src/shared_math/traits.rs
```diff
@@ -74,18 +75,19 @@ pub trait New {
...
-pub trait PrimeField:
+pub trait FiniteField:
     Clone
     + Eq
     + Display
     + Serialize
     + DeserializeOwned
     + PartialEq
     + Debug
-    + IdentityValues
```
sshine, line 88: It seems that this trait can now be removed, too?

twenty-first/src/shared_math/x_field_element.rs
```diff
@@ -242,44 +224,52 @@ impl Display for XFieldElement {
...
+        }
     }
 
-    #[inline]
-    fn ring_one(&self) -> Self {
-        Self::ring_one()
+    fn is_one(&self) -> bool {
+        self.coefficients[0].is_one()
+            & self.coefficients[1].is_zero()
+            & self.coefficients[2].is_zero()
```
sshine, line 267:
```suggestion
        self.coefficients[0].is_one()
            && self.coefficients[1].is_zero()
            && self.coefficients[2].is_zero()
```

twenty-first/src/shared_math/rescue_prime_regular.rs
```diff
@@ -0,0 +1,1479 @@
...
+    }
+
+    /// hash_varlen
+    /// Hash an arbitrary number of field elements. Takes care of
+    /// padding.
+    pub fn hash_varlen(input: &[BFieldElement]) -> [BFieldElement; 5] {
+        let mut sponge = RescuePrimeRegularState::new();
+
+        // pad input
+        let mut padded_input = input.to_vec();
```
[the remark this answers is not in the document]
aszepieniec: There is a benchmark for `rescue_prime_regular` which also runs a benchmark for the `varlen` variant, so it is possible to answer this question. However, re-writing the code to avoid the extra clone seems rather tricky. Care to suggest something?

twenty-first/src/shared_math/traits.rs
aszepieniec: That's weird, I did not find this line in my version. 7603b93b does delete the (unimplemented) trait though.

twenty-first/src/shared_math/x_field_element.rs
aszepieniec: Rust overloads the operator `BitAnd`, *i.e.*, `&`, to support the usual meaning when applied to bools.

I think the differences between `&` and `&&` in rust 
 - `&&` is only a logical operator whereas `&` also has other meanings – bitwise and, borrow, *etc.*
 - `&&` is lazy whereas `&` is not.

Field elements are typically used for cryptography where timing attacks, which can result from lazy evaluation, can be fatal. I would suggest using `&` by default. If the application allows, another method using `&&` can be written for faster average run time. But I'm not sure that can even be achieved because the lazy evaluation will throw the branch predictor into disarray.

sshine: > I would suggest using `&` by default.

Do you mean
1) in low-level arithmetic like `add()`, `mul()`, `is_one()`, `mod_pow_u32()`,
1) in high-level arithmetic like `Polynomial<T>` and `MPolynomial<T>`,
1) in other timing-critical code than arithmetic, e.g. in hash functions, and/or
1) in general?

I believe we can divide all arithmetic optimizations we've made so far into "using algebraic properties, but possibly or deliberately constant-time", "deliberately using short-circuiting" (especially in the case of polynomial arithmetic), and "using memory efficiently". In short, if `&` vs. `&&` makes a difference here, we might want to (a) re-visit a large number of optimizations made so far, since some might have a larger and more obvious signature than `is_one()`, and (b) evaluate the [`CMOVcc` instructions for x86_64](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DEArgoKkl9ZATwDKjdAGFUtEywZ7HAGTwNMAHLuAEaYxCDSAA6oCoR2DC5uHnrRsbYCvv5BLKHh0laYNvFCBEzEBInunlyWmNbpDCVlBJmBIWERlqXllck15s2t2bkRAJSWqCbEyOwcAPRzANQATACkAKzYAGySiwC0KxvYAMzLi6vLAEKLXKsaAIKRJsGLaAzmi8o7IIsmO%2BfHAAiiw0qio4KoAH0IVCNHCNJD4bdjpc7o9nq8BB9lFxlgAOH4mXF4gHA0Ew6EQxHw6kaZGoh5op4vKgMRZMdDoCDqX47UiLACehJ2o32q2O2F5u1WAHYGfdForFvQCIsFO5CcTSYtVOyFL8tRdrgK9Qb8eL5UrFhB1Sxzst1p9iaKmPq/pI0bLAZ6HszFqz2Zzoa6CNzhdJBeHRXtxZL3ec5WirSrFiwTKrbZr8drda6zSSjYLTUTzSik0q8FQ1e4AZKcdnZZarUrbWKgU7S03FV7y4rW3n3Z6Zd7Gb6MQGOehIQB3PwEMJUOq0MNS/lCqXR2NShNdxYLdn264vcXApj7a2Rc/BUa95WYVXcmqvLiik86gB0qAAbgvaKhZ0YkLqsEEBfLsBwCjeZYPMm96Bto2oaCYpzvtOxBMJEkR%2BMAQHPBAyBcMWpxQbuqhcKh6GYdhuEgRyCEDiKQ4jvcTLjmyabLjy7prlGbZxv8ja3imkTEBgJg2FmBbtrmbpagAVEWA7Ehat4QCJYk2Pajr1niLpujsTE%2BuiLLsSYtCQsQmDoOJmBcOsACcK7cZGG58dugkwUqwmidZEn5jmxbyYpsmdrelm%2BbZDlqT54kEDejLDkZfoBhxFlWTZkIsLIwCoGwxACk5fIue6m4Su5iaeYq3kaQQkkBUp2YKSaDV4iplWLOFGVZYIOV5QV6m%2BXFhmjvcCxLPcCi2pgiwEAgTCqrqbyxOY%2BqoFW/BTIspx7MEhCLF%2BYgmJgCg/Ew/LBPyyD8ugIBomNd37gcZ5vKUfj6qcd4hmqeDAAwlZ4KIgiLLtBAKKQD1LAc6CYoITBvVtZxZR8sS/f9gOqiDCjvhDEM6oseD6sEokANaMPjgioJixCWQo0QMPgRj7Ydx2mgoCD/myoR/tOaJLaq3gAPIAOrYAASpCACy9xCAA0uGiFglSMJtSxjLLMcGyXH4tB%2BJgEBiNOTACgoN7rMxAadTMdmOVxzpufGHkKl5cFMMe0nFgZ0FO1VcHINDb7crWW4ih7Hpe7BqrINq%2BHQwOxGhyrEeLP77Z%2B0HZWnCrt77mgLBPPO7IvFDFrA4UTBmNN0MsEbpeLMAlnzWEM1zWyrv8i36D0Is0TzoIeBiL89O/v%2BQlwRABB5xo/IE5CJhD8Qr7tq7n4/sQVDc9RwEQOgJGj6qE%2BRIRb4HxoFEYVhgFb4LIvi1LsuLApEAz3PWDEKHoy7yNVo57l%2BfTVHCk1ibA%2BgcZAJdZoE3xvqYAJgyiGHnFZRYDBUD70pi/YePN2opgPmcAOUdxROAIQjaMrws7tX3JOGamA85MzcCzQw0NLIECmGyWa00aZmQICXUIogK6LEwEwZACBFhrWbtNecNDq4mlCL8S8RBFj/ExuDchSwO5dx7owWwA9vwYL3taDhtALBQMhDohe2oD7kVMevf81FJzjzzssT%2B3sOrHU4WfKigE7HXzFpLaWMsH7WhnqY9%2B8UWKJRGhcDW6wtYMB1v4fWtBDbG1NsxPm/o2SW0wJlbKuUwgFVtviUq/FpQVWcfuIQmB2GYEXJZBgMxC46P9CQA8WAFDIGIHgSIDQRFVjYU02gG9GZ4DzvQNgsMGjYywXBVQtB6r6TDruFMqhhEB11LGYOkg9JSkTs7B8Z1%2BGL2BDMleGDbGcm5LMghRDiJkOcSmN2p53EXxwlvM86yM6OKeZvPC00GKbNuUnCAZBXiHJ1AgE5a9BkvLwteHZipiBfMvnhJCKE0Ln2%2BSBKOcdHHv2GvcDg4xaCcHWLwTwHAtCkFQJwUWZgMyTGmNNSJPBSC1XJQS8YxMQDrCnkSjgkhSWaF4FSjgvATpT1ZVocYcBYBIFzlhegZAKD4V/nQDoxAuBbGOFPGghiwgnQgMEQVpBdrMHypwZlucxkEAFrEoUbLSBYGrkYcQ9r8CWSKD%2BE69rMCqEKOmWYzK5x1CNTrImZQBQuCwEaggHSWDmoJXwAwwAFAADU8CYGnALSIjB40yEECIMQ7ApB5vkEoNQRrdA1AMEYEAphzD6DwMEE6kBxioC6fEL1ewnAiK6XsegP5aAnmOKKuohQGgOHpr0aopAfC62GB0SQ9lSCpDiAIKdKQYiroYEMdoeQl0FCKAIJoPRXBVD0Aehox6Whzt3REfd3QKinr6F0QYN6cgLvsuMBQ9KZh6BjZU%2BNhLiUCvtcKutBAo7kU1e%2BDQ1oaX1sWLgQgzSmX8hcHnVVb8mWjF4BK9lpBOXcv0JwflpAyUUuFaKkA4rBUf2IxwZYIGKOcFw7R8Yq9Yj2EkEAA%3D%3D%3D) that the efficient prime-field arithmetic uses ([`csel` on aarch64](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DEArgoKkl9ZATwDKjdAGFUtEywZ7HAGTwNMAHLuAEaYxCAAHBqkAA6oCoR2DC5uHnpxCbYCvv5BLKHhUZaY1lkMQgRMxAQp7p5cxaVJFVUEOYEhYXrmrbVpDT3V7XkFEgCUlqgmxMjsHAD08wDUAEwApACs2ABskksAtKub2ADMK0trKwBCS1xrGgCCMSbBS2gM5kvKuyBLJrsXE4AESWGlUVAhVAA%2BpDoRp4RooQi7icrvcni83gJPsouCsIr8THiIoCQWDYTDIUiEdSNCi0Y90c9XlQGEsmOh0BB1H9dqQlgBPQm7MYHNYnbC8vZrADsDIeS0VS3oBCWCnchOJpKWqnZCj%2BWsuNwFeoN%2BPF8qVSwg6pYFxWGy%2BxNFTH1/0k6NlQM9j2ZS1Z7M5MNdBG5wukgvDov24sl7oucvRVpVSxYJlVts1%2BO1utdZpJRsFpqJ5tRSaVeCoavcgMluOzsstVqVtrFwKdpabiq95cVrbz7s9Mu9jN9mIDHPQUIA7n4CGEqCVaGGpfyhVLo7GpQmu0tFuz7TdXuKQUwDtaYufgmNe8rMKruQ03lxRSedQA6VAANwXtFQs6MKF1WCCBvj2Q4BRvMtHmTe9A20bUNBMM532nYgmBiGI/GAICXggZAuGLM4oN3VQuFQ9DMOw3CQI5BCBxFIcRweJlxzZNNlx5d01yjNs4wBRtbxTGJiAwEwbCzAt21zN0tQAKiLAdiQtW8IBEsSbHtR16wiF03V2JifQxFl2JMWgoWITB0HEzAuA2ABOFduMjDc%2BO3QSYKVYTROsiT8xzYt5MU2TO1vSzfNshy1J88SCBvRlhyMv0Aw4iyrJsqEWFkYBUDYYgBScvkXPdTcJXcxNPMVbyNIISSAqU7MFJNBqIhUyqlnCjKssEHK8oK9TfLiwzRweRZlgeBRbUwJYCAQJhVV1d4EnMfVUCrfhpiWM59mCQgli/MQTEwBRfiYflgn5ZB%2BXQEB0TGu790OM93kqPx9TOO8QzVPBgAYSs8FEQQll2ggFFIB7lkOdAsUEJg3q284ss%2BBJfv%2BwHVRBhR3whiGdSWPB9WCUSAGtGHxwRUCxYhLIUOIGHwIx9sO47TQUBB/zZUI/2ndEltVbwAHkAHVsAAJShABZB4hAAaXDRDwSpWE2pYxkVhOTYrj8Wg/EwCAxGnJgBQUG8NmYgNOtmOzHK4503PjDyFS8uCmGPaTiwM6CnaquDkGht9uVrLcRQ9j0vdg1VkG1fDoYHYjQ5ViOln99s/aDsqzhV299zQFhnnndlXihi1gcwUQzGm6GWCN0ulmASz5rCGa5rZV3%2BRb9B6CWOJ50EPAxD%2Benf3/IS4IgAg8%2BifGFChEwh%2BIV921dz8f2IKhueo4CIHQEjR9VCeYkIt8D40CiMKwwCt8FkXxal2WlgUiACdn%2BfQ7GXeRqtHPcvz6ao4U9YWwPqHGQCXWaBNp51xMFUQw84rJLAYKgfelM55YDXtzPeM087nADlHcUTh8EI2jG8LO7V9yThmpgPOTM3As0MNDSyBBphslmtNGmZkCAl1COXJQSxMBMGQAgJYa1m7TXnNQ6uJpQh/EvEQJYAJMbgzIcsDuXce6MFsAPb8w8ebtRTBAdhtALDTyhNohe2oD7kTMevf81FJzj2wR/b2HVjocLPlRQC9jr5i0ltLGWD9rTPzMW/eKLFEojUuBrDYWsGA638PrWghtjam2Ynzf0bJLaYEytlXKYQCq23xKVfi0oKrOP3EITAbDMCLksgwWYhdtH%2BhIAeLAChkDEDwDEMowiqysKabQDejM8B53oGwWGZRsZ6LgqoWg9V9Jh13CmVQQiA66ljMHSQekpSJ2dg%2BM6fDF4ghmSvHRdjOTclmfgwhxFSHOJTG7U87iL44S3medZGcVhjCeZvPC00GKbNuUnAxl1Dk6gQCc9BtjL54WvDsxUxBvnQpAkhFCaFz4/JAlHOOny37DQeBwCYtBOAbF4J4DgWhSCoE4KLMwGYpgzGmpEngpBarkoJRMYmIANjRCJRwSQpLNC8CpRwXgJ1oisq0BMOAsAkC5ywvQMgFB8I/zoF0YgXBtgnGiDQIxYQToQGCIK0gu1mD5U4My3OYyCAC1iUKNlpAsDVyMOIe1%2BBLI2DwD%2BE69rMCqDLumOYzK5wlCNTrImVQBQuCwEaggHSWDmoJXwAwwAFAADU8CYGnALGIjAE0yEECIMQ7ApD5vkEoNQRrdANAMEYEAphzD6DwMEE6kAJioC6Ukb1%2BwnDCK6fsegP5aAnhOAcfYlRiDABdlUQRux9hz2Jog6cDB%2B1%2BBMKofYv0TCipKGXMoDh6Z9HqKQHwusRhdAaBkRIAhD3pHiFehgwxOjhAGDuj1AgWjVBvS%2Bpo77x1tFPU%2B7of6v2WD/Y%2B/I56JgKHpbMPQsbKkJsJcSgV9rhX1oIFHcimr3waGtDShtSxcCEGaUy/kLg86quIPaE4L5eASvZaQTl3L9CcH5aQMlFLhWipAOKwV78WMcBWChzjnA6N8YmKvBI9hJBAA)).

twenty-first/src/shared_math/rescue_prime_regular.rs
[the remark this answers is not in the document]
sshine: I did something quirky once for the `RescuePrimeXlix` struct that apparently conflicted with the way we intend to do padding, so the version that copies less hasn't been in production for three months. I'll try and dig up an old diff. For now, let's assume that this is efficient enough and we might have room to improve.

twenty-first/src/shared_math/x_field_element.rs
aszepieniec: Comparison is the most obvious candidate for a function having a sensitive time signature. I would argue that we should avoid variable-time running times for low-level arithmetic like `add` and `mul` but I'm not sure that is a matter of urgent concern.

For high-level arithmetic including `Polynomial` and `MPolynomial` – specifically when the size is not fixed at compile-time – striving for a constant-time implementation is just silly because the running time has to depend on the size of the data.

twenty-first/src/util_types/simple_hasher.rs
```diff
@@ -120,110 +134,112 @@ pub trait Hasher: Sized + Send + Sync + Clone {
...
 // node count as a u128 makes this possible
-impl ToDigest<Blake3Hash> for u128 {
-    fn to_digest(&self) -> Blake3Hash {
-        (*self).into()
+
+impl Hashable<u8> for u128 {
+    fn to_sequence(&self) -> Vec<u8> {
+        (0..8)
+            .map(|i| ((*self >> (8 * i)) & 0xff) as u8)
+            .collect::<Vec<u8>>()
```
Sword-Smith, line 141: A `u128` is 16 bytes, not 8.

twenty-first/src/util_types/simple_hasher.rs
```diff
@@ -120,110 +134,112 @@ pub trait Hasher: Sized + Send + Sync + Clone {
...
-        ]
+impl Hashable<u8> for XFieldElement {
+    fn to_sequence(&self) -> Vec<u8> {
+        let mut array: Vec<u8> = Vec::with_capacity(3 * 8);
+        for i in 0..3 {
+            for j in 0..8 {
+                array.push(((u64::from(self.coefficients[i]) >> (j * 8)) & 0xff) as u8);
+            }
+        }
+        array
```
Sword-Smith, line 153: Two elements representing the same value can hash to two different digests with this code. Maybe convert to `canonical_representation` first?

twenty-first/src/util_types/simple_hasher.rs
```diff
@@ -120,110 +134,112 @@ pub trait Hasher: Sized + Send + Sync + Clone {
...
 }
 
-impl ToDigest<Blake3Hash> for Blake3Hash {
-    fn to_digest(&self) -> Blake3Hash {
-        *self
+impl Hashable<u8> for BFieldElement {
+    fn to_sequence(&self) -> Vec<u8> {
+        (0..8)
+            .map(|i| ((u64::from(*self) >> (i * 8)) & 0xff) as u8)
+            .collect::<Vec<u8>>()
```
Sword-Smith, line 161: Same problem as for `Hashable<u8>` for `XFieldElement`.

twenty-first/src/util_types/simple_hasher.rs
```diff
@@ -337,55 +332,66 @@ impl RescuePrimeXlix<RP_DEFAULT_WIDTH> {
...
+            "Cannot sample pseudo-uniform BFieldElements from less than 8 bytes."
+        );
+        let mut integer = 0u64;
+        digest[0..8].iter().for_each(|d| {
+            integer *= 256u64;
+            integer += *d as u64;
+        });
+        BFieldElement::new(integer)
+    }
+}
```
Sword-Smith, line 336: Prone to collisions since `BFieldElement::new(x) == BFieldElement::new(x + Q)`.

aszepieniec:
> We might be able to get rid of hash_with_salts. It was used for salted Merkle trees, but we might be able to provide the salts another way or let the caller handle the salts such that the salted Merkle trees can be scrapped alltogether.

[0ab54e5](https://github.com/Neptune-Crypto/twenty-first/pull/25/commits/0ab54e5b0165262501c032c7a21fdbf3b77ae62b) drops `hash_with_salts`; please review.

I don't think we need salted Merkle trees anywhere in Neptune but it's a solid module and it would be a shame to drop it.

twenty-first/src/util_types/simple_hasher.rs
aszepieniec: `u64::from( BFieldElement ) -> u64`
converts the BFieldElement to canonical representation before casting.

twenty-first/src/util_types/simple_hasher.rs
aszepieniec: The issue here is not collisions but non-uniformity. However, it _is_ an issue impacting security with a corresponding attack.

aszepieniec: Resolved in [0da29bd](https://github.com/Neptune-Crypto/twenty-first/pull/25/commits/0da29bd2934d6b85ddf652ea936f34d59bea560c).

twenty-first/src/shared_math/x_field_element.rs
sshine: Let's look at the timing for the low-level stuff subsequently, then.

Consider the `&` → `&&` suggestion retracted.

sshine:
> make sure that he's happy with the new hasher trait

I'm very happy with this.

aszepieniec:
Ready from my end.

Sword-Smith:
> Ready from my end.

No objections here.

twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -269,6 +253,12 @@ impl From<u32> for BFieldElement {
     }
 }
 
+impl From<u64> for BFieldElement {
+    fn from(value: u64) -> Self {
+        Self(value)
+    }
+}
```
sshine, line 260: This behavior is unexpected.

From the [`std::convert::From` documentation](https://doc.rust-lang.org/std/convert/trait.From.html):

> **Note: This trait must not fail.** The From trait is intended for perfect conversions. If the conversion can fail or is not perfect, use [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html).

It is not generally true that `u64`s convert to `BFieldElement`s, since any `u64` $\geq P\not\in$ `BFieldElement`. 

By claiming that e.g. `BFieldElement::QUOTIENT.into() == BFieldElement::zero()` is expectable, this would justify `impl From<u64> for u32` or any other lossy conversion as implicitly modular. I think there is risk associated with making lossy conversion the default, since it might encourage (either accidentally or unknowingly) a conversion to happen when the caller does not assume that their `u64` contains a non-canonical `BFieldElement` value.

aszepieniec: Adding context without arguing (my opinion is not qualified enough yet):

I added `From<u64> for BFieldElement` even though `BFieldElement::new` already takes u64s: `u64 -> BFieldElement`. The rationale was that `new` does the modular reduction if necessary. But sometimes, like when reading hardcoded constants from the program code that are known to be BFieldElements and just happen to be encoded as u64s, you want to create the BFieldElement object directly and bypass the validity test and possible modular reduction. The way I read the documentation was that `from<u64> for BFieldElement` was the right way to do that.

PR #27, opened 2022-09-12
Implement flush.

Ulrik-dk:


Sword-Smith:
Thanks for the PR. Let's remove all other `flush` than the explicit ones while we're at it.

twenty-first/src/util_types/mmr/archival_mmr.rs
```diff
@@ -145,6 +145,11 @@ where
         }
     }
 
+    // Flush the databases
```
Sword-Smith, line 148: Let's make this singular, not plural.

Sword-Smith (changes requested):

Sword-Smith (approved):

PR #30, opened 2022-09-13
README: Add release protocol

sshine:
This set of steps describes how a new release is made.

This protocol ensures that backwards-breaking changes are not released with a "major-compatible" version number, causing projects that depend on this library to break unexpectedly.

It also ensures better ability to track what changes were made. This helps the discovery of regressions and gives a better overview of each release.

---

Note that I've applied this protocol for the releases 0.1.5 and 0.2.0. They appear here:

- https://github.com/Neptune-Crypto/twenty-first/tags
- https://crates.io/crates/twenty-first/versions

Note also that the releases 0.1.0, 0.1.1, 0.1.2 and 0.1.3 don't (yet) have git tags.


Sword-Smith:
Nice. Thanks for making this!

PR #35, opened 2022-09-19
Add `inverse_or_zero()` to {B, X}FieldElement

### LINKED ISSUE #33: Add `inverse_or_zero()` to {B, X}FieldElement
In Triton VM, it is a common pattern to invert a field element if it the element is non-zero. Instead of writing the same code all the time, supply method `mul_inverse_or_zero()`.

jan-ferdinand:
Fix #33

### DIFF
--- a/twenty-first/src/shared_math/traits.rs
+++ b/twenty-first/src/shared_math/traits.rs
@@ -16,9 +16,17 @@ where
 // TODO: Assert if we're risking inverting 0 at any point.
 pub trait Inverse
 where
-    Self: Sized,
+    Self: Sized + Zero,
 {
     fn inverse(&self) -> Self;
+
+    fn inverse_or_zero(&self) -> Self {
+        if self.is_zero() {
+            Self::zero()
+        } else {
+            self.inverse()
+        }
+    }
 }
 
 pub trait PrimitiveRootOfUnity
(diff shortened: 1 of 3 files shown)


twenty-first/src/shared_math/traits.rs
```diff
@@ -16,9 +16,17 @@ where
 // TODO: Assert if we're risking inverting 0 at any point.
 pub trait Inverse
 where
-    Self: Sized,
+    Self: Sized + Zero,
```
jan-ferdinand, line 19: I'm not sure this is a breaking change, requiring a version bump of “major.” Let me know.

sshine: Since all `impl Inverse` are for types that have `Zero`, we won't experience it that way, but hypothetical subscribers to the trait alone would experience it as breaking, yes. I'd mark this as breaking, but also: You don't need to release a new version of twenty-first yet because of this. We have a handful of small changes we'd like to push as a result of merging #25 onto other codebases. Let's schedule a version bump.

sshine (approved):

Sword-Smith:
Looks good to me. We used to have a batch-inversion function that did the above. Maybe you would want that functionality back if this is something you are doing a lot?

Sword-Smith (approved):
Good tests. I especially like that you made sure to hit the `zero` case :)

PR #38, opened 2022-09-26
Move STARKs to separate crates in Cargo workspace

sshine:
BrainSTARK and the Rescue-Prime STARK are proof-of-concept zk-STARK engines built with cryptography from the twenty-first library. The tutorials for this code is found at:

- https://aszepieniec.github.io/stark-anatomy/
- https://aszepieniec.github.io/stark-brainfuck/

Both BrainSTARK and the Rescue-Prime STARK have lived as modules in twenty-first until they are now being moved into their own crates under a Cargo workspace. This means both STARKs are continuously compiled, tested and benchmarked against twenty-first, but are not distributed in the twenty-first crate at crates.io:

https://crates.io/crates/twenty-first

---

**Remark:** This is a breaking change to the twenty-first 0.2.0 API. A new crate release is delayed.


stark-brainfuck/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "stark-brainfuck"
+version = "0.2.0"
+authors = ["Triton Software AG"]
+edition = "2021"
+
+license = "GPL-2.0"
+description = "A Turing-complete zk-STARK engine consisting of a virtual machine, prover and verifier."
+homepage = "https://github.com/Neptune-Crypto/twenty-first"
```
jan-ferdinand, line 9: Would it make sense to link to the tutorial here?

stark-brainfuck/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+version = "0.3"
+features = ["html_reports"]
+
+[dev-dependencies.cargo-husky]
+version = "1"
+default-features = false
+features = ["precommit-hook", "run-cargo-clippy", "run-cargo-fmt"]
+
+[dependencies]
+twenty-first = { path = "../twenty-first" }
```
jan-ferdinand, line 30: Why not the crates.io crate?

jan-ferdinand (approved):

stark-brainfuck/Cargo.toml
sshine: Because stark-brainfuck functions as a test suite for twenty-first. If we depend on the crates.io version, running `make` or `cargo ...` will not trigger failure when a branch makes a breaking change. We would like to know that twenty-first diverged from being compatible with stark-brainfuck before it hits crates.io.

Related: Making twenty-first a path dependency only will prevent stark-brainfuck from being uploaded to crates.io. If we should want to do that, it is possible to [specify both `path` and `version`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations). This requires bumping the version number in another location, though:

```suggestion
twenty-first = { path = "../twenty-first", version = "0.2.0" }
```

stark-brainfuck/Cargo.toml
sshine: Yes!

PR #40, opened 2022-10-07
Implement `formal_derivative` refactor `xgcd`.

einar-triton:


twenty-first/src/shared_math/polynomial.rs
```diff
@@ -2754,4 +2804,86 @@ mod test_polynomials {
...
+    fn differentiate_zero() {
+        let elm = BFieldElement::new(0);
+        let p = Polynomial::new_const(elm);
+        let q = p.formal_derivative();
+
+        assert!(q.is_zero());
+        assert_eq!(q.degree(), -1)
+    }
+    #[test]
+
```
jan-ferdinand, line 2818: Clippy was happy with this? :fearful: 
```suggestion

    #[test]
```

jan-ferdinand (approved):
I petition to generalize `xgcd` from `BFieldElement` to `FiniteFieldElement`, but don't veto merging this PR.

Sword-Smith:
> I petition to generalize `xgcd` from `BFieldElement` to `FiniteFieldElement`, but don't veto merging this PR.

Sounds reasonably to me, but then we would need tests of other cases than `BFieldElement` I would say.

jan-ferdinand:
I concur. Writing tests for something we don't currently need is something better left for a slow day.

sshine:
Did you see that `xgcd()` was already implemented generically?

https://github.com/Neptune-Crypto/twenty-first/blob/master/twenty-first/src/shared_math/other.rs#L123-L150

jan-ferdinand:
I did not know that existed. Trying to switch it, it turns out that `other::xgcd()` requires trait bounds `Zero` and `One`, among others, both of which are not implemented for `Polynomial<FFElem>`. The `XFieldElement::xgcd()` does work for `Polynomial<BFieldElement>` – ultimately, this is all I currently need. The fact that this method lives on `XFieldElement` just struck me as weird, hence I asked for a refactor.

Sword-Smith:
I'm implementing `Zero` and `One` for `Polynomial` now. This is possible after @aszepieniec removed the deprecated "flexible" prime field type.

Sword-Smith:
I tried using the generic `xgcd` instead of the one added here, but there seems to be some normalization going on in the newly added `xgcd` that does not happen in the generic `xgcd` implementation found in `other`. This lack of normalization means the test added here does not succeed when using the generic `xgcd`.

 Also: The one defined for `Polynomial<BFieldElement>` only does one long division whereas the generic one will do two long divisions on polynomials. The clue is that the quotient and remainder is found simultaneously in the newly added `xgcd` version.

Sword-Smith:
Updated with upstream, so `Polynomial` now implements the `Zero` and `One` traits.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -1074,10 +1067,67 @@ impl<PFElem: FiniteField> Sub for Polynomial<PFElem> {
...
+        mut y: Polynomial<BFieldElement>,
+    ) -> (
+        Polynomial<BFieldElement>,
+        Polynomial<BFieldElement>,
+        Polynomial<BFieldElement>,
+    ) {
+        let mut a_factor = Polynomial::new(vec![BFieldElement::one()]);
+        let mut a1 = Polynomial::new(vec![BFieldElement::zero()]);
+        let mut b_factor = Polynomial::new(vec![BFieldElement::zero()]);
+        let mut b1 = Polynomial::new(vec![BFieldElement::one()]);
```
jan-ferdinand, line 1083:
```suggestion
impl<FFelem: FiniteField> Polynomial<FFelem> {
    pub fn xgcd(
        mut x: Polynomial<FFelem>,
        mut y: Polynomial<FFelem>,
    ) -> (
        Polynomial<FFelem>,
        Polynomial<FFelem>,
        Polynomial<FFelem>,
    ) {
        let mut a_factor = Polynomial::new(vec![FFelem::one()]);
        let mut a1 = Polynomial::new(vec![FFelem::zero()]);
        let mut b_factor = Polynomial::new(vec![FFelem::zero()]);
        let mut b1 = Polynomial::new(vec![FFelem::one()]);
```
No reason to limit this implementation to polynomials over $\mathbb{F}_p$ for one specific $p$.

jan-ferdinand (changes requested):
At some point, we should rename `PFelem` to `FFelem` or `GFelem` where it makes sense. But that might be for another commit, and probably will not need review.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -1074,10 +1067,67 @@ impl<PFElem: FiniteField> Sub for Polynomial<PFElem> {
...
+
+        while !y.is_zero() {
+            let (quotient, remainder): (Polynomial<BFieldElement>, Polynomial<BFieldElement>) =
```
jan-ferdinand, line 1086:
```suggestion
            let (quotient, remainder): (Polynomial<FFelem>, Polynomial<FFelem>) =
```

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -1074,10 +1067,67 @@ impl<PFElem: FiniteField> Sub for Polynomial<PFElem> {
...
         degree_raw(&self.coefficients)
     }
+
+    pub fn formal_derivative(&self) -> Self {
+        let coefficients = self
+            .clone()
+            .coefficients
+            .iter()
+            .enumerate()
+            .map(|(i, &coefficient)| PFElem::new_from_usize(&coefficient, i) * coefficient)
```
jan-ferdinand, line 1125: I suggest changing the trait for `new_from_usize` to not take `&self` an argument.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: Done. I'm OK with this suggestion now that I have added tests for both B field and X field polynomials.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: I agree. But we can do that in a separate commit/PR.

Sword-Smith:
New PR is up. The xgcd for polynomials is now normalized (leading coefficient is one) but it uses the generic method. Let me hear your thoughts.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -2420,10 +2450,44 @@ mod test_polynomials {
...
+    pub fn xgcd_b_field_pol_test() {
+        for _ in 0..100 {
+            let x: Polynomial<BFieldElement> = gen_polynomial_non_zero();
+            let y: Polynomial<BFieldElement> = gen_polynomial_non_zero();
+            let (gcd, a, b): (
+                Polynomial<BFieldElement>,
+                Polynomial<BFieldElement>,
+                Polynomial<BFieldElement>,
+            ) = Polynomial::xgcd(x.clone(), y.clone());
+            assert!(gcd.is_one());
```
jan-ferdinand, line 2463: I think the probability of this test failing even though the gcd was computed correctly, i.e., two random polynomials of degree at most 40 sharing a factor, is smaller than $\frac{40\cdot40}{p} \approx 8.7\cdot 10^{-17}$, which is probably acceptable. Still, something in me itches…

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -1074,10 +1067,47 @@ impl<PFElem: FiniteField> Sub for Polynomial<PFElem> {
     }
 }
 
+impl<PFElem: FiniteField> Polynomial<PFElem> {
+    // Division in 𝔽_p[X], not 𝔽_{p^e} ≅ 𝔽[X]/p(x).
+    /// Run the extended Euclidean algorithm on the polynomial ring. Normalizes the
+    /// result the inverse to a leading coefficient of one.
```
jan-ferdinand, line 1073:
```suggestion
    /// Extended Euclidean algorithm with polynomials. Computes the greatest
    /// common divisor `gcd` as a monic polynomial, as well as the corresponding
    /// Bézout coefficients `a` and `b`, satisfying `gcd = a·x + b·y`.
```

jan-ferdinand (approved):
Left two minor remarks regarding comments, and one regarding tests that might trigger false negatives. Other than that, looking good, I'd say.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -1074,10 +1067,47 @@ impl<PFElem: FiniteField> Sub for Polynomial<PFElem> {
     }
 }
 
+impl<PFElem: FiniteField> Polynomial<PFElem> {
+    // Division in 𝔽_p[X], not 𝔽_{p^e} ≅ 𝔽[X]/p(x).
```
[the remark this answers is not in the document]
Sword-Smith: OK.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: I'm not sure my polynomial understanding suffices to improve this test. Feel free to change it yourself though.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: Done

PR #41, opened 2022-10-08
Thv/speed up xfe ntt

Sword-Smith:
This speeds up NTT by ~~60 %~~ 80-100 %. But should the interface be changed as I suggest here, or is there another solution that's preferable?


sshine:
Could it be as simple as adding `+ Mul<Output = BFieldElement>` on the NTT's `PFElem: FiniteField` and adding

```rust
impl Mul for XFieldElement {
    type Output = BFieldElement;

    ...
}
```

in addition to the existing `impl Mul for XFieldElement`?

Sword-Smith:
> Could it be as simple as adding `+ Mul<Output = BFieldElement>` on the NTT's `PFElem: FiniteField` and adding
> 
> ```rust
> impl Mul for XFieldElement {
>     type Output = BFieldElement;
> 
>     ...
> }
> ```
> 
> in addition to the existing `impl Mul for XFieldElement`?
No, the return type of the new mul implementation is `XFieldElement`.

Sword-Smith:
> I'm sorry, my copy-pasting went a little fast. I meant, of course:
> 
> ```rust
> impl Mul for XFieldElement {
>     type Output = XFieldElement;
> 
>     fn mul(&self, rhs: BFieldElement) -> Self::Output { ... }
> }
> ```
> 
> since the `Mul` trait supports asymmetric operand types.
> 
> Alternatively, if this does not work because of overlapping trait instances, name it something that does not break abstraction in the trait, i.e. finite fields do not know of b-field elements.

We still need the `mul` implementation for multiplying two X-field elements.

sshine:
Sorry, I'm giving this half attention as I write. Let me clarify what's possible:

Instead of having one `impl Mul for XFieldElement` where the implicit type parameter `Rhs = XFieldElement`, have two:

```rust
// This is equivalent to the `impl Mul for XFieldElement` that we currently have
impl Mul<XFieldElement> for XFieldElement {
    type Output = Self;

    fn mul(self, rhs: XFieldElement) -> Self::Output {
        todo!()
    }
}

// This allows the right-hand side of `*` to be a `BFieldElement`
impl Mul<BFieldElement> for XFieldElement {
    type Output = Self;

    fn mul(self, rhs: BFieldElement) -> Self::Output {
        todo!()
    }
}
```

Now, you can add `+ Mul<BFieldElement, Output = Self>` as a supertrait to the `FiniteField` trait instead of inventing `MulBFieldElement`.

tl;dr: Your trait already exists, it's called `Mul`.

Sword-Smith:
I added the `Mul<BFieldElement>` to `XFieldElement`. I like this solution. 

Users of `twenty-first` should take note though: You will get a speedup with this change but only if you use it right: You can now multiply an X field element with a B field element but only if the B field element is on the RHS of the multiplication sign. `x =* b` where `x` is an X field element and `b` a B field element will also give the speedup these commit introduce.

This change increases speed by 80-100 % depending on the size. For a domain of 2^23, the time is reduced from 1.8s to 1.0s on my fast Core i9 machine.

stark-brainfuck/src/xfri.rs
```diff
@@ -727,7 +721,13 @@ mod xfri_tests {
...
             get_x_field_fri_test_object(subgroup_order, expansion_factor, colinearity_check_count);
         let mut proof_stream: StarkProofStream<Hasher> = StarkProofStream::default();
-        let subgroup = fri.domain.omega.get_cyclic_group_elements(None);
+        let subgroup: Vec<XFieldElement> = fri
+            .domain
+            .omega
+            .get_cyclic_group_elements(None)
+            .iter()
+            .map(|x| x.lift())
+            .collect();
```
sshine, line 730: This allocates the entire FRI domain twice in memory.

Maybe this does not matter, since it's tutorial code.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -816,7 +560,270 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
...
+            coefficients: vec![element],
+        }
+    }
+
+    // FIXME: Can be done with traits instead of explicitly mentioning B and X.
+    // Thor does not agree that this is a good path to venture down
+    pub fn lift_b_x(b_poly: &Polynomial<BFieldElement>) -> Polynomial<XFieldElement> {
+        let x_field_coefficients = b_poly.coefficients.iter().map(|b| b.lift()).collect();
+        Polynomial::new(x_field_coefficients)
+    }
```
sshine, line 583: Do we even still use this? I see no callers.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -816,7 +560,270 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
...
+        while !self.coefficients.is_empty() && self.coefficients.last().unwrap().is_zero() {
+            self.coefficients.pop();
+        }
+    }
+
+    pub fn zero() -> Self {
+        Self {
+            coefficients: vec![],
+        }
+    }
```
sshine, line 595: Move to `impl Zero for ...` instance.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -816,7 +560,270 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
...
+
+    pub fn is_zero(&self) -> bool {
+        self.coefficients.is_empty() || self.coefficients.iter().all(|x| x.is_zero())
+    }
+
+    pub fn is_one(&self) -> bool {
+        self.degree() == 0 && self.coefficients[0].is_one()
+    }
```
sshine, line 609: Move to `impl Zero for ...` and `impl One for ...` instances.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -116,75 +117,12 @@ impl<PFElem: FiniteField> PartialEq for Polynomial<PFElem> {
...
-    }
-
-    pub fn leading_coefficient(&self) -> Option<PFElem> {
-        match self.degree() {
-            -1 => None,
-            n => Some(self.coefficients[n as usize]),
-        }
-    }
-
+impl<PFElem: FiniteField + MulAssign<BFieldElement>> Polynomial<PFElem> {
```
sshine, line 120: Consider moving this to a `where` now that it's grown.

```suggestion
impl<PFElem> Polynomial<PFElem>
where
    PFElem: FiniteField + MulAssign<BFieldElement>
{
```

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -197,135 +135,67 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
...
             return Self::zero();
         }
 
+        // A benchmark run on sword_smith's PC revealed that
+        // `fast_square` was faster when the input size exceeds
+        // a length of 64.
         let squared_coefficient_len = self.degree() as usize * 2 + 1;
+        if squared_coefficient_len > 64 {
+            return self.fast_square();
+        }
```
sshine, line 187: Scientists say... ;-)

Maybe we can do a benchmark here.

This seems like a thing that could be quite machine-dependent.

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -197,135 +135,67 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
...
 
+        // TODO: Review.
         for i in 0..self.coefficients.len() {
             let ci = self.coefficients[i];
             squared_coefficients[2 * i] += ci * ci;
 
-            // TODO: Review.
             for j in i + 1..self.coefficients.len() {
                 let cj = self.coefficients[j];
                 squared_coefficients[i + j] += two * ci * cj;
```
sshine, line 201: - [x] Reviewed.

*(What are we assessing here, exactly?)*

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -336,177 +206,43 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
...
     }
-}
 
-impl<PFElem: FiniteField> Polynomial<PFElem> {
-    pub fn are_colinear(points: &[(PFElem, PFElem)]) -> bool {
-        if points.len() < 3 {
-            println!("Too few points received. Got: {} points", points.len());
-            return false;
+    #[must_use]
+    pub fn fast_mod_pow(&self, pow: BigInt) -> Self {
```
sshine, line 211: Why `pow: BigInt` as opposed to `PFElem`, `u64`, `u32` or `u8`?

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -816,7 +560,270 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
...
+
+        for i in 0..self.coefficients.len() {
+            let ci = self.coefficients[i];
+            squared_coefficients[2 * i] += ci * ci;
+
+            // TODO: Review.
+            for j in i + 1..self.coefficients.len() {
+                let cj = self.coefficients[j];
+                squared_coefficients[i + j] += two * ci * cj;
+            }
```
sshine, line 763: - [x] Reviewed.

*(What are we assessing here, exactly?)*

twenty-first/src/shared_math/x_field_element.rs
```diff
@@ -305,7 +304,22 @@ impl Add for XFieldElement {
 
 */
 
-impl Mul for XFieldElement {
+impl Mul<BFieldElement> for XFieldElement {
```
sshine, line 307:
```suggestion
/// Scalar multiplication of `XFieldElement`
impl Mul<BFieldElement> for XFieldElement {
```

twenty-first/src/shared_math/polynomial.rs
```diff
@@ -730,7 +470,11 @@ impl<PFElem: FiniteField> Polynomial<PFElem> {
     }
 
     /// The inverse of `fast_coset_evaluate`
-    pub fn fast_coset_interpolate(offset: &PFElem, generator: PFElem, values: &[PFElem]) -> Self {
+    pub fn fast_coset_interpolate(
+        offset: &BFieldElement,
+        generator: BFieldElement,
+        values: &[PFElem],
+    ) -> Self {
```
sshine, line 477: I suppose this means that we don't generally `fast_coset_interpolate`.

sshine (approved):
This optimization comes at a cost of generality.

I know that we've avoided an extension-field trait.

This PR makes me want to reconsider that.

stark-brainfuck/src/xfri.rs
Sword-Smith: I agree that it doesn't matter, not right now at least.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: Why not?

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: One data point is better than zero.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: No idea.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: This is just code I moved around, since the MulAssign<BFieldElement> trait requirement had to be added to some methods.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: This is just code I moved around, since the MulAssign<BFieldElement> trait requirement had to be added to some methods.

twenty-first/src/shared_math/x_field_element.rs
Sword-Smith: I disagree with this comment. I'm not sure exactly what "scalar" refers to.

twenty-first/src/shared_math/polynomial.rs
Sword-Smith: I think we only run it over `XFieldElement` codewords, but this code will also work for `BFieldElement` codewords. We always use an offset and a generator from the B field though. The faster NTT needs to know that the generator is a `BFieldElement` now, that change in NTT propagates up to many of the NTT-based polynomial functions.

PR #48, opened 2022-10-23
Drain padded input instead of re-allocating vector

sshine:
This results in a ~19% speedup of `RescuePrimeRegular::hash_varlen()`.

Running `cargo bench varlen` before and after:

```
Benchmarking rescue_prime_regular/hash_varlen/RescuePrimeRegular / Hash Variable Length/16384:
time before:  [28.763 ms 28.812 ms 28.866 ms]
time after:   [23.233 ms 23.252 ms 23.277 ms]
change:      [-19.462%  -19.296%  -19.133%] (p = 0.00 < 0.05)
```

### DIFF
--- a/twenty-first/src/shared_math/rescue_prime_regular.rs
+++ b/twenty-first/src/shared_math/rescue_prime_regular.rs
@@ -1024,7 +1024,7 @@ impl RescuePrimeRegular {
             {
                 *sponge_state_element += input_element.to_owned();
             }
-            padded_input = padded_input[RATE..].to_vec();
+            padded_input.drain(..RATE);
             Self::xlix(&mut sponge);
         }
 


sshine:
Another unnecessary `.to_vec()` is made when copying the entire `input` into `padded_input`:

```rust
pub fn hash_varlen(input: &[BFieldElement]) -> [BFieldElement; 5] {
    let mut sponge = RescuePrimeRegularState::new();

    // pad input
    let mut padded_input = input.to_vec();
    padded_input.push(BFieldElement::one());
    while padded_input.len() % RATE != 0 {
        padded_input.push(BFieldElement::zero());
    }

    ...
```

Since the padded input is also copied into the sponge `RATE` elements at a time, padding can happen then.

This would make Rescue-Prime use a constant amount of memory for each `hash*()` operation.

sshine:
A more realistic assessment of the speedup depends on the expected input sizes for `::hash_varlen()`.

PR #51, opened 2022-11-04
ignore zero-coefficients for computing polynomials total degree

### LINKED ISSUE #50: Error in `MPolynomial::degree`
This method seems to wrongly count terms whose coefficient is zero towards the degree of a multivariate polynomial. It shouldn't.

https://github.com/Neptune-Crypto/twenty-first/blob/master/twenty-first/src/shared_math/mpolynomial.rs#L1057-L1071

jan-ferdinand:
Fix #50.

### DIFF
--- a/stark-brainfuck/src/vm.rs
+++ b/stark-brainfuck/src/vm.rs
@@ -162,7 +162,7 @@ pub fn compile(source_code: &str) -> Option<Vec<BFieldElement>> {
     let mut program: Vec<BFieldElement> = vec![];
     let mut stack: Vec<usize> = vec![];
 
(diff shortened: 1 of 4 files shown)


Sword-Smith:
Nice. That was fast :)

Sword-Smith:
LGTM

Sword-Smith (approved):
Good tests as always :)

jan-ferdinand:
> Good tests as always :)

Could be better, but I couldn't think of anything fast enough. :relieved:

Sword-Smith:
> > Good tests as always :)
> 
> Could be better, but I couldn't think of anything fast enough. 😌

There are a lot of tests that touch `degree` for mpols already. No need to exaggerate for this cornercase.

PR #52, opened 2022-11-04
Allow hashing of multivariate polynomials

Sword-Smith:
This allows us to put the multivariate polynomials into HashSet which will
be required for an efficient implementation of circuits. Through this HashSet
we can check if we are inserting a value that has already been encountered
and if so then use that already encountered value instead.

### DIFF
--- a/twenty-first/src/shared_math/mpolynomial.rs
+++ b/twenty-first/src/shared_math/mpolynomial.rs
@@ -10,6 +10,7 @@ use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::error::Error;
 use std::fmt::{Debug, Display};
+use std::hash::Hash;
 use std::iter::Sum;
 use std::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub};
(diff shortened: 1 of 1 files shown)


twenty-first/src/shared_math/mpolynomial.rs
```diff
@@ -2370,4 +2387,89 @@ mod test_mpolynomials {
...
+            );
+
+            // Digest must change if mpoly changes (non-zero coefficient term)
+            rnd_mvpoly
+                .coefficients
+                .insert(vec![2; variable_count], BFieldElement::one());
+            let mut hasher3 = DefaultHasher::new();
+            rnd_mvpoly.hash(&mut hasher3);
+            let hash3 = hasher3.finish();
+            assert_ne!(hash0, hash3, "hashing must change if mpoly changes");
```
jan-ferdinand, line 2424: minor syntactical note
```suggestion
            assert_ne!(hash0, hash3, "hash digest must change if mpoly changes");
```

jan-ferdinand (approved):

PR #54, opened 2022-11-13
Arithmetic overloading

sshine:
- `pub const EXTENSION_DEGREE: usize = 3;`
- Add the following instances:
  - `impl Add<XFieldElement> for XFieldElement` (original)
  - `impl Add<BFieldElement> for XFieldElement` (new)
  - `impl Add<XFieldElement> for BFieldElement` (new)
  - `impl Mul<XFieldElement> for XFieldElement` (original)
  - `impl Mul<BFieldElement> for XFieldElement` (new)
  - `impl Mul<XFieldElement> for BFieldElement` (new)
  - `impl Sub<XFieldElement> for XFieldElement` (original)
  - `impl Sub<BFieldElement> for XFieldElement` (new)
  - `impl Sub<XFieldElement> for BFieldElement` (new)

Assert that they corresponding to lifting the BFieldElements.

**Note:** This only warrants a minor bump.

### DIFF
--- a/twenty-first/src/shared_math/x_field_element.rs
+++ b/twenty-first/src/shared_math/x_field_element.rs
@@ -12,9 +12,11 @@ use crate::shared_math::polynomial::Polynomial;
 use crate::shared_math::traits::{CyclicGroupGenerator, FiniteField, ModPowU32, ModPowU64, New};
 use crate::shared_math::traits::{FromVecu8, Inverse, PrimitiveRootOfUnity};
 
+pub const EXTENSION_DEGREE: usize = 3;
+
 #[derive(Debug, PartialEq, Eq, Copy, Clone, Hash, Serialize, Deserialize)]
 pub struct XFieldElement {
-    pub coefficients: [BFieldElement; 3],
+    pub coefficients: [BFieldElement; EXTENSION_DEGREE],
 }
 
 impl Default for XFieldElement {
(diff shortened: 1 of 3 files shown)


twenty-first/src/shared_math/x_field_element.rs
```diff
@@ -12,9 +12,11 @@ use crate::shared_math::polynomial::Polynomial;
 use crate::shared_math::traits::{CyclicGroupGenerator, FiniteField, ModPowU32, ModPowU64, New};
 use crate::shared_math::traits::{FromVecu8, Inverse, PrimitiveRootOfUnity};
 
+pub const EXTENSION_SIZE: usize = 3;
```
aszepieniec, line 15:
```suggestion
pub const EXTENSION_DEGREE: usize = 3;
```

aszepieniec: https://en.wikipedia.org/wiki/Degree_of_a_field_extension

Sword-Smith:
I like the tests. Maybe remove the FIXME comment? And why does the operator order for sub need to be changed?

sshine:
- [x] Maybe remove the FIXME comment? 

> And why does the operator order for sub need to be changed?

To have a consistent ordering instead of an arbitrary one.

The comment for `impl Mul<XFieldElement> for XFieldElement` was resting above the comment for `impl Mul<BFieldElement> for XFieldElement`, making it slightly out of context. I figured that if I'm going to make the comments make sense, I could also improve the reading order.

Sword-Smith:
I think this warrants a major bump as some type inference will not longer work. So it's actually a breaking change.

PR #56, opened 2022-11-16
decouple univariate polynomial's evaluation domain from parent field

jan-ferdinand:
Make methods relating to evaluation and interpolation of univariate polynomials generic over the domain using generic type `GF`, which has to be compatible with `FF`. This de-couples the domain over which interpolation or evaluation is to happen from the field the polynomial is defined over.

For example, consider $f \in \mathbb{Q}[x]$ to be evaluated in $S \subseteq \mathbb{Z}$.

Note that `FF` and `GF` do not have to be distinct.

### DIFF
--- a/twenty-first/src/shared_math/polynomial.rs
+++ b/twenty-first/src/shared_math/polynomial.rs
@@ -133,6 +133,12 @@ impl<FF: FiniteField> PartialEq for Polynomial<FF> {
 
 impl<FF: FiniteField> Eq for Polynomial<FF> {}
 
+/// Univariate polynomial over finite field `FF`. Methods relating to evaluation and interpolation
+/// are generic over the domain using `GF`, which has to be compatible with `FF`. This allows
+/// de-coupling the domain over which interpolation or evaluation is to happen from the field the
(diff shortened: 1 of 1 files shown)


twenty-first/src/shared_math/polynomial.rs
```diff
@@ -322,7 +328,14 @@ where
...
 
     // domain: polynomial roots
-    pub fn fast_zerofier(domain: &[FF], primitive_root: &BFieldElement, root_order: usize) -> Self {
+    pub fn fast_zerofier<GF>(
+        domain: &[GF],
+        primitive_root: &BFieldElement,
+        root_order: usize,
+    ) -> Self
+    where
+        GF: FiniteField + Into<FF>,
```
sshine, line 337: Adding a type parameter without a default is a breaking change, so `0.x+1.0`.

I'd probably mention that `GF: Into<FF>` means projecting `GF` into a superset in all of the cases of this PR.

Maybe that's a comment worth having at the top rather than at each `<GF: FiniteField>`.

sshine (approved):

PR #59, opened 2022-11-18
Add getter function for values of U32s<N>

Ulrik-dk:


Ulrik-dk:
I now see that the convention seems to be a getter function, instead of exposed values. I will update this PR to that convention.

sshine (commented):
As discussed, let's make this an [`impl AsRef<[u32; N]>`](https://doc.rust-lang.org/std/convert/trait.AsRef.html).

sshine (approved):

PR #64, opened 2022-11-23
Emojihash trait

sshine:
- Move .emojihash() to Emojihash trait
- Create generic impls for vectors, arrays, slices
- Use generic impls for XFieldElement and Digest


einar-triton (commented):
neat!

PR #66, opened 2022-12-12
Add `::sample_weights()` to AlgebraicHasher, implement with `::hash_pair()`

sshine:
sample_weights:
- Move from Triton VM's Stark. It is highly generic, and having it here as a public function means that it's exposed. (This will help when making the mirrored translation in tasm, testing equivalence.)

sample_indices:
- Change order and name of arguments.
- Take `seed: &Digest` for consistency.
- Make call sites more verbose as to which argument is which.

get_n_hash_rounds:
- Change implementation so that it uses `::hash_pair()` instead of `::hash_varlen()`, since the latter is not available from within Triton VM.
- Encode the counter as the top stack element because this presumably gives the shortest tasm code.


twenty-first/src/util_types/algebraic_hasher.rs
```diff
@@ -48,26 +48,48 @@ pub trait AlgebraicHasher: Clone + Send + Sync {
...
+            .map(|counter: usize| {
+                let counter = Digest::new([
+                    BFieldElement::new(counter as u64),
+                    BFieldElement::zero(),
+                    BFieldElement::zero(),
+                    BFieldElement::zero(),
+                    BFieldElement::zero(),
+                ]);
+                Self::hash_pair(&counter, seed)
+            })
```
jan-ferdinand, line 92: @aszepieniec, could I get your input here?

jan-ferdinand (approved):

twenty-first/src/util_types/algebraic_hasher.rs
sshine: Technically, this only allows $2^{64} - 2^{32}$ samples, after which it starts repeating.

A more general solution could be to encode the counter as a `U32s<N>`.

Considering we need `security_level` (160) samples in the practical case, the current should be sufficient.

I've added the following assertion:

```rust
assert!(count <= BFieldElement::MAX as usize);
```

twenty-first/src/shared_math/x_field_element.rs
```diff
@@ -131,7 +131,8 @@ impl XFieldElement {
 
     pub fn sample(digest: &Digest) -> Self {
         let elements = digest.values();
-        XFieldElement::new([elements[0], elements[1], elements[2]])
+        XFieldElement::new([elements[2], elements[3], elements[4]])
+        // XFieldElement::new([elements[0], elements[1], elements[2]])
```
sshine, line 135: This change currently results in a clockcycle improvement of $5729 \rightarrow 5209$ for `sample_weights(130)`.

```rust
// Current: _ s4 s3 s2 s1 s0 addr i _w4 _w3 w2 w1 w0

// If we instead of picking sample := [w0, w1, w2], we pick sample := [w2, w3, w4]... we save 8 clockcycles per sample:

pop pop                        // _ s4 s3 s2 s1 s0 addr i w4 w3 w2

// This works with no prior change to `XFieldElement::sample()`:

// swap3 pop                   // _ s4 s3 s2 s1 s0 addr i _w4 w0 w2 w1
// swap3 pop                   // _ s4 s3 s2 s1 s0 addr i w1 w0 w2
// swap2                       // _ s4 s3 s2 s1 s0 addr i w2 w0 w1
// swap1                       // _ s4 s3 s2 s1 s0 addr i w2 w1 w0
```

PR #68, opened 2022-12-14
Switch to Montgomery representation on B field

Sword-Smith:
This leads to a 20-30 % reduction in time for NTT on my P17 machine and a 15-25 % reduction in runtime with an AMD Threadripper 5995.

The internal representation of the B field element is a bit more complicated now, so some of the tests had to be changed as well as they shouldn't access or have opinions about the internal representation, only the results of various calculations.

### DIFF
--- a/twenty-first/src/shared_math/b_field_element.rs
+++ b/twenty-first/src/shared_math/b_field_element.rs
@@ -477,6 +477,7 @@ mod b_prime_field_element_test {
     use crate::shared_math::polynomial::Polynomial;
     use itertools::izip;
     use proptest::prelude::*;
+    use rand::thread_rng;
 
     #[test]
     fn display_test() {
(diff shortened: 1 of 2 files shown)


jan-ferdinand (approved):

Sword-Smith:
Since this removes a few public functions from the B field module, it gives rise to a new major version.

sshine (approved):

twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -160,55 +160,31 @@ impl BFieldElement {
...
-        let tmp1 = tmp0.wrapping_sub(Self::LOWER_MASK * (under as u64));
-
-        // compute c * 2^32 - c; this is guaranteed not to underflow
-        let tmp2 = (c << 32) - c;
-
-        // add temp values and return the result; because each of the temp may be up to 64 bits,
-        // we need to handle potential overflow
-        let (result, over) = tmp1.overflowing_add(tmp2);
-        result.wrapping_add(Self::LOWER_MASK * (over as u64))
+    const fn mod_reduce(x: u128) -> u64 {
```
aszepieniec, line 172: Since this method does not compute modular reduction, I would suggest not calling it that.

The [source material](https://eprint.iacr.org/2022/274.pdf) calls it `montyred`, which is presumably short for "Montgomery reduction", which is distinct from modular reduction modulo $p$.

PR #69, opened 2022-12-21
make MDS matrices & round constants compile-time known `BFieldElement`s

jan-ferdinand:
Hopefully, this resolves the performance degradation of hashing.


Sword-Smith:
I'm having trouble overcoming the hash/Merkle tree slowdown introduced by the conversion to Montgomery representation of the `BFieldElement`. The only idea I have at the moment is to try to declare the MDS and round constants directly from the u64 representation of their Montgomery representations.

Sword-Smith:
Worth considering, from ##rust on IRC Freenode:

```IRC
14:32 < sword_smith> If I declare a big structure (a matrix in my case) as const and I build it using calls to `const` functions, is 
                     the data structure guaranteed to be known at compile time? 
14:32 < sword_smith> Am I guaranteed to not use runtime cycles on computing this matrix?
14:44 < polychromata> sword_smith: const /items/ are evaluated only at compiletime, yes
14:44 < polychromata> calls to const /functions/ may or may not be (you can think of them as const /compatible/ functions)
14:45 < polychromata> but if the end result is going in a const item then you know it has to be done at compiletime
...
15:45 < sword_smith> polychromata: Thanks. That's also what I expected. But that means that I can't understand why our hash function is 
                     slower than it was before.
15:52 < sword_smith> I changed something from `const u64` to `const BFieldElement(u64)` and it got a lot slower. Maybe the reason is 
                     that it can do more at compile time when it's a primitive type?
15:52 < sword_smith> "it" being rustc with full optimization.
15:56 < bertptrs> the optimizer sometimes gets confused with newtypes, this is a known problem

```

Sword-Smith:
The main problem is that the benchmark `rescue_prime_regular/hash_varlen/RescuePrimeRegular / Hash Variable Length/16384` has slowed down from 25.7ms to 67.9ms on my P17 machine, or equivalently on other machines. So the change from old representation to Montgomery representation has slowed down hashing by a factor 2.5.

Sword-Smith:
IRC suggests using Flamegraph to see where the time is lost. Let's try that!

Sword-Smith:
I managed to identify the relevant chunk of code for the problem

This is our old `add` implementation. The one we used before changing to Montogomery representation.

That weird if/branch in the middle is responsible for the speedup in pre-Montgomery code (e.g. `v0.8.0`). Without it `rescue_prime_regular/hash_varlen/RescuePrimeRegular / Hash Variable Length/16384` takes 57ms on our reference machine. With it, the same benchmark takes 25.6ms. But other benchmarks are slower with that magical chunk of code.

I'm leaning towards accepting the slowdown we have introduced with Montgomery and then fixing either through `tip5` or by being inspired by Winterfell's RP implementation.


```rust
fn add(self, other: Self) -> Self {
        let (result, overflow) = self.0.overflowing_add(other.0);
        let mut val = result.wrapping_sub(Self::QUOTIENT * (overflow as u64));

        // For some reason, this `if` codeblock improves NTT runtime by ~10 % and
        // Rescue prime calculations with up to 45 % for `hash_pair`. I think it has
        // something to do with a compiler optimization but I actually don't
        // understand why this speedup occurs.
        if val > Self::MAX {
            val -= Self::QUOTIENT;
        }

        Self(val)
    }
```

jan-ferdinand:
> Let's try [flamegraphs]!

I think this is a good idea generally speaking.

int-e:
#70 is my investigation into this; I've managed to reclaim the semi-mysterious speedup for `add` in this patch: https://github.com/Neptune-Crypto/twenty-first/pull/70/commits/585fc10993e15d98cc5c42f393847c0d801d0f73

Sword-Smith:
Superseeded by @int-e's valuable contribution in #70.

PR #72, opened 2023-01-05
Tip5 Hash Function

aszepieniec:
This PR introduces the Tip5 hash function. Since it is the only extant implementation, this implementation is also the reference implementation.


Sword-Smith (changes requested):
Karatsuba multiplication:
- You add Karatsube multiplication for polynomials, but you don't seem to be using it anywhere. Why then add it?
- If Karatsuba should stay, I would prefer to have it better tested for corner-cases. Like 1 * 1, 0 * 0, f(x) * 0, 0 * f(x) etc.

In Tip5:
The `Tip5State::new` function can be made `const` allowing it to be called at compile time. If that option is there, I think it should be exploited. This requires line 23 i `tip5.rs` to be changed to 
```rust
state: [BFieldElement::new(0); STATE_SIZE],
```
The `new` function for `Tip5` could also be implemented as `const` if you precalculate the round constants instead of calculating them dynamically. The function `offset_fermat_cube_map` can be made `const` by just adding that keyword.

I guess exposing `BFieldElement`'s `montyred` makes sense because this allows you to use the left shift operation instead of multiplication?

You're making some pretty radical changes to `add` in `BFieldElement`. Are you sure that's wise? Have you benchmarked it to verify that you're not introducing a slowdown? The change doesn't seem to introduce a difference in the `tip5` hash functions but your change *does* slow down the `RescuePrimeRegular` hash function meaning that this PR shows a too large speedup. I still see at least a factor 8 speedup across the board from this PR though.

aszepieniec:
> Karatsuba multiplication:
>
>    You add Karatsube multiplication for polynomials, but you don't seem to be using it anywhere. Why then add it?
>    If Karatsuba should stay, I would prefer to have it better tested for corner-cases. Like 1 * 1, 0 * 0, f(x) * 0, 0 * f(x) etc.

Karatsuba-multiplication was a potential strategy for fast MDS multiplication, but in the end NTT is faster. Theoretically there should be a regime where Karatsuba is the fastest multiplication algorithm but as it works with the Polynomial wrapper it is at a disadvantage relative to NTT-based multiplication, which operates directly on the vectors of coefficients. I don't know if there is any degree for which this current Karatsuba outperforms NTT-based multiplication.

> In Tip5:
> The Tip5State::new function can be made const allowing it to be called at compile time. If that option is there, I think it > should be exploited. This requires line 23 i tip5.rs to be changed to
> 
> `state: [BFieldElement::new(0); STATE_SIZE],`
> 
> The new function for Tip5 could also be implemented as const if you precalculate the round constants instead of calculating them dynamically. The function offset_fermat_cube_map can be made const by just adding that keyword.

`Tip5State::new` can be made into a const and this has been done. However, `Tip5::new` cannot be made const, or at least not in its current form. The reason is that the coefficients are determined via Blake3, for which there is no const implementation.

It is possible to hardcode the constants as consts (and same for the lookup table) but that defeats the purpose of having a `Tip5` struct to begin with.

> I guess exposing BFieldElement's montyred makes sense because this allows you to use the left shift operation instead of multiplication?

Correct. Although, unfortunately it seems that bitshifting does not improve performance over multiplying by an arbitrary field element.

> `BFieldElement`
> You're making some pretty radical changes to add in BFieldElement. Are you sure that's wise? Have you benchmarked it to verify that you're not introducing a slowdown? The change doesn't seem to introduce a difference in the tip5 hash functions but your change does slow down the RescuePrimeRegular hash function meaning that this PR shows a too large speedup. I still see at least a factor 8 speedup across the board from this PR though.

No! I expect that this change arises from accepting the wrong change in some or other rebase. Let me revert back to the same code that's on master.

Sword-Smith (approved):

PR #77, opened 2023-01-08
Minor Improvements

aszepieniec:
- Add modular reduction function
- Add fixed tests
- Rename const variable `QUOTIENT` to `P`


Sword-Smith (changes requested):
I'm not sure what `mod_reduce` function is doing here. The rest I'm happy with.

PR #81, opened 2023-01-12
Hashing: Increase OUTPUT_LENGTH from 5 to 10

sshine:
AlgebraicHasher:
 - hash_op: The interface now holds a function that is essentially hash_10, but 1) does not mention the number 10, and 2) lives on the interface.
 - hash_pair: Since hash_op is available on the interface, it can be defined generally instead of in the trait instance for each hash function.
 - INPUT_LENGTH: This is currently 10. Now it has a variable!
 - OUTPUT_LENGTH: This is currently 5, but will be increased to 10.

Digest:
 - DIGEST_LENGTH: Move to rescue_prime_digest because it's a part of the interface for digests rather than any concrete hash function implementation. (There are currently 3 competing, and they all adhere to the same digest length, but redefine the constant.)

### DIFF
--- a/stark-rescue-prime/src/stark_rp.rs
+++ b/stark-rescue-prime/src/stark_rp.rs
@@ -9,6 +9,7 @@ use std::convert::TryInto;
 use std::error::Error;
 use std::fmt;
 use std::iter::zip;
+use twenty_first::shared_math::rescue_prime_digest::DIGEST_LENGTH;
 
 use twenty_first::shared_math::b_field_element::BFieldElement;
 use twenty_first::shared_math::fri::Fri;
--- a/twenty-first/benches/rescue_prime_optimized.rs
+++ b/twenty-first/benches/rescue_prime_optimized.rs
@@ -4,7 +4,8 @@ use rand::RngCore;
 use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use twenty_first::shared_math::b_field_element::BFieldElement;
 use twenty_first::shared_math::other::random_elements;
-use twenty_first::shared_math::rescue_prime_optimized::{RescuePrimeOptimized, DIGEST_LENGTH};
+use twenty_first::shared_math::rescue_prime_optimized::RescuePrimeOptimized;
+use twenty_first::util_types::algebraic_hasher::OUTPUT_LENGTH;
 
 fn bench_10(c: &mut Criterion) {
     let mut group = c.benchmark_group("rescue_prime_optimized/hash_10");
@@ -59,7 +60,7 @@ fn bench_parallel(c: &mut Criterion) {
                 elements
                     .par_iter()
                     .map(RescuePrimeOptimized::hash_10)
-                    .collect::<Vec<[BFieldElement; DIGEST_LENGTH]>>()
+                    .collect::<Vec<[BFieldElement; OUTPUT_LENGTH]>>()
             });
         },
     );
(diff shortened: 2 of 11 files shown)


sshine:
After these changes, in order for `impl AlgebraicHasher for RescuePrimeRegular`, it looks like:

```rust
impl AlgebraicHasher for RescuePrimeRegular {
    fn hash_op(input: &[BFieldElement; INPUT_LENGTH]) -> [BFieldElement; OUTPUT_LENGTH] {
        RescuePrimeRegular::hash_10(input)
    }

    fn hash_slice(elements: &[BFieldElement]) -> Digest {
        Digest::new(RescuePrimeRegular::hash_varlen(elements))
    }
}
```

Just like `hash_op` corresponds to the `hash` instruction, I would further like to modify the trait so that instead of saying how it `hash_slice`s, it should say how it `squeeze`s and `absorb`s statefully, so that `hash_slice` can be defined as a trait fn defined in terms of `squeeze` and `absorb`. This makes the AlgebraicHasher trait very close to what's possible in Triton VM, so that a dual tasm implementation is more trivial.

sshine:
Compliance tests rely on expected values that only contain the first 5 elements of the sponge. In order to reach proper compliance, they need to be re-generated with 10 elements. Currently a bunch of 0s are in place of the missing elements in order for the tests to fail for the right reasons.

```
failures:
    shared_math::rescue_prime_regular::rescue_prime_regular_tests::test_compliance
    shared_math::tip5::tip5_tests::hash10_test_vectors
```

jan-ferdinand:
I don't concur with the changes proposed here. For `hash_pair` to work efficiently in (for example) the context of generating Merkle trees, the length of the output should be half the length of the input – the length of the digest should stay 5.

sshine:
> For hash_pair to work efficiently in (for example) the context of generating Merkle trees, the length of the output should be half the length of the input – the length of the digest should stay 5.

This is an interface change; it shouldn't affect performance, no extra allocations are being made.

`DIGEST_LENGTH` is still 5.

Perhaps I can emphasise the motivation here:

For those things that need equivalent implementations inside Triton VM and in Rust, the closer the primitives are inside Triton VM and outside (in twenty-first, mostly), the more alike these implementations will be. I've identified three use-cases for hashing:

- Merkle trees, for which `hash_pair` and the `hash` instruction work fine already; it is not necessary to increase `OUTPUT_LENGTH` beyond `DIGEST_LENGTH` in this scenario, since the `hash` instruction was specifically built for the purpose of effectively hashing Merkle trees.
- Sample `XFieldElement` weights / `u32` indices, for which the `hash` instruction can only sample one weight per call to `hash` (using 3 of 5 field elements): This is wasteful considering the `RATE` is 10. A better solution makes 3 sample weights out of one call to `hash`, and an even better solution makes 3.333..., although it remains to be tested if that's actually faster wrt. clock cycles. With only `hash` available in Triton VM, efficient sampling is not possible without increasing `OUTPUT_LENGTH` for `hash_10` and the `hash` instruction.
- Hashing large objects efficiently: Since yesterday I understand that extending the instruction set with `absorb` and `squeeze` instructions makes it possible to efficiently hash large objects without exposing more than `RATE` elements. Before this is possible, a larger output window for the `hash` instructions would be crucial, but still sub-optimal.

Performance-wise you still pop the same amount of elements off when hashing Merkle trees inside Triton VM after `OUTPUT_LENGTH` is increased from 5 to 10. The values popped off just wouldn't be zeros, and can be used for other purposes.

An alternative path to efficient sampling is to use `absorb` and `squeeze` assuming `squeeze` has a bigger output size than 5. Maybe that's what you are proposing?

sshine:
tl;dr:
- Keeping the `hash` instruction to `DIGEST_LENGTH` limits its use, but makes its interface less confusing.
- Switching sampling from counter-mode to squeezing (with `squeeze` and `absorb` in Triton VM) provides the necessary flexibility.

PR #82, opened 2023-01-12
Deprecate STARK tutorial libraries by moving into stark-shared crate

sshine:
This repository is a multi-crate repository, of which only the crate twenty-first is released on crates.io. The remaining crates contain tutorial code. With this change, code only used by tutorials is moved into a separate crate shared between the two tutorial STARKs.

The shared code that is being moved:
 - ProofStream
 - Fri
 - FriDomain
 - StarkVerifyError
 - MerkleProofError


sshine:
One note: It seems that `stark-brainfuck` uses both the shared FRI's `FriDomain` and its own.

There is some possibility for cleaning that up which this pull request preserves perfectly.

PR #84, opened 2023-01-20
Replace old AlgebraicHasher with AlgebraicHasher: SpongeHasher

sshine:
- hash_slice() becomes hash_varlen()
- impl SpongeHasher + AlgebraicHasher for blake3::Hasher


sshine:
While this commit passes twenty-first, it incurs breaking changes in triton-vm that involve rewriting `.prove()` and `.verify()` for both Stark and FRI. This PR should not be merged unless this is resolved, or it is estimated that placing such a breaking change on twenty-first master is acceptable.

PR #86, opened 2023-01-26
make domain separation for hash10 apply to all capacity elements

aszepieniec:


jan-ferdinand (approved):
I don't find the new `new()` particularly beautiful but cannot improve on it given the limitations of `const fn`.

sshine (approved):

PR #88, opened 2023-02-01
Fix tip5

Nashtare:
The current version of `split_and_lookup` computes the correct S-Box but does not update the mutable element passed as argument. As a consequence, the 4 first state elements are never updated when calling `sbox_layer`.

This PR fixes it and regenerates test vectors accordingly.


aszepieniec (approved):
Excellent catch, thank you!

I was expecting this might adversely affect performance, but bizarrely it seems to *improve* performance:

image

PR #91, opened 2023-02-15
Faster MDS matrix

aszepieniec:


hybrid.sage
```diff
@@ -0,0 +1,687 @@
+load("mds.sage")
```
jan-ferdinand, line 1: This file is not included in the PR.

jan-ferdinand (approved):
I can't claim to have understood every method in this PR. :relieved:

aszepieniec:
To avoid bloating history, I merged only the final `tip5.rs` directly into master.

PR #94, opened 2023-02-24
add algebraic prng trait

aszepieniec:
This trait provides a home for the function `sample_indices` to live.


twenty-first/src/shared_math/rescue_prime_regular.rs
```diff
@@ -1038,6 +1038,7 @@ impl AlgebraicHasher for RescuePrimeRegular {
 }
 
 impl SpongeHasher for RescuePrimeRegular {
+    const RATE: usize = RATE;
```
sshine, line 1041: Adding `RATE` to the trait seems fine, but references to `RATE` in the functions of this trait need to be `Self::RATE` unless they should refer to one of the other `RATE`s that exist (each hash function has its own `pub const RATE: usize`, and there is a shared one in algebraic_hasher.rs.

twenty-first/src/util_types/algebraic_prng.rs
```diff
@@ -0,0 +1,37 @@
...
+        let num_squeezes =
+            (num_elements * 3 + <H as SpongeHasher>::RATE - 1) / <H as SpongeHasher>::RATE;
+        (0..num_squeezes)
+            .into_iter()
+            .flat_map(|_| <H as SpongeHasher>::squeeze(state))
+            .collect_vec()
+            .chunks(3)
+            .map(|ch| XFieldElement::new(ch.try_into().unwrap()))
+            .collect_vec()
+    }
```
sshine, line 26: The reason `sample_indices()` and `sample_weights()` were recently removed from twenty-first is because the way weights are sampled is highly specific to Triton VM; that is, whatever way is most convenient to sample weights inside the VM should affect how we sample them in STARK/FRI (which are also in Triton VM).

I had originally placed them on this trait because I wanted to bundle everything related to hashing in one place. Ferdinand convinced me that it's probably better to have them (or at least `sample_weights()`) in Triton VM. (You could similarly argue that `squeeze` and `absorb` are quite coupled to Triton VM as well.)

If I understand correctly, we have two independent subscribers to `sample_indices()`, but only one subscriber to `sample_weights()`, which is Triton VM. If that is true, we might consider only having `sample_indices()` in twenty-first.

twenty-first/src/util_types/algebraic_prng.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    fn sample_indices(state: &mut Self::PrngState, max: u32, num_indices: usize) -> Vec<u32> {
+        let num_squeezes =
+            (num_indices + <H as SpongeHasher>::RATE - 1) / <H as SpongeHasher>::RATE;
+        (0..num_squeezes)
+            .into_iter()
+            .flat_map(|_| <H as SpongeHasher>::squeeze(state))
+            .map(|b| (b.value() % max as u64) as u32)
+            .collect()
+    }
```
sshine, line 36: Why not just have this method live on `AlgebraicHasher`?

What's the idea behind `AlgebraicPrng`?

sshine (commented):
If I understand the picture right, I would recommend:
- Move `sample_indices()` to `AlgebraicHasher` and avoid the `AlgebraicPrng` trait
- Keep `sample_weights()` in Triton VM assuming it only gets used there.
- If `RATE` should be an associated type of `AlgebraicHasher`, ~remove the `pub const RATE: usize` in algebraic_hasher.rs and replace occurrences of `RATE` on the trait and its implementations with `Self::RATE`~.

**Update:** It occurred to me that the reason I hadn't made `RATE` an associated type is that it's being used to determine the length of arrays. It is a current limitation in the Rust compiler that `generic parameters may not be used in const operations`.

twenty-first/src/util_types/algebraic_prng.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl<H: SpongeHasher> AlgebraicPrng for H {
+    type PrngState = H::SpongeState;
+
+    fn sample_xfield(state: &mut Self::PrngState, num_elements: usize) -> Vec<XFieldElement> {
+        let num_squeezes =
+            (num_elements * 3 + <H as SpongeHasher>::RATE - 1) / <H as SpongeHasher>::RATE;
+        (0..num_squeezes)
+            .into_iter()
```
sshine, line 20: Ranges are already iterators.

```suggestion
```

twenty-first/src/util_types/algebraic_prng.rs
```diff
@@ -0,0 +1,37 @@
...
+            .flat_map(|_| <H as SpongeHasher>::squeeze(state))
+            .collect_vec()
+            .chunks(3)
+            .map(|ch| XFieldElement::new(ch.try_into().unwrap()))
+            .collect_vec()
+    }
+
+    fn sample_indices(state: &mut Self::PrngState, max: u32, num_indices: usize) -> Vec<u32> {
```
sshine, line 28: If `sample_indices()` does not require that `max` is a power of 2, we will lack an algorithm for sampling indices on Triton VM. I assume that our only use-case of sample indices that aren't powers of 2 will be deprecated (it's in the current FRI, but it's not necessary).

twenty-first/src/util_types/algebraic_prng.rs
```diff
@@ -0,0 +1,37 @@
+use itertools::Itertools;
+
+use crate::shared_math::x_field_element::XFieldElement;
+
+use super::algebraic_hasher::SpongeHasher;
+
+pub trait AlgebraicPrng {
+    type PrngState;
+    fn sample_xfield(state: &mut Self::PrngState, num_elements: usize) -> Vec<XFieldElement>;
+    fn sample_indices(state: &mut Self::PrngState, max: u32, num_elements: usize) -> Vec<u32>;
```
sshine, line 10: Calling the parameter `max: u32` makes it sound like it's an inclusive upper bound, which it is not.

Calling it `upper_bound: u32` leaves it ambiguous, but not indicating towards the wrong interpretation.

twenty-first/src/util_types/algebraic_prng.rs
sshine: I would assume that `sample_indices(sponge, max, num_indices).len() == num_indices`, but the current implementation rounds up to the nearest multiple of `RATE`. While this is equivalent to the behavior on Triton VM with the `squeeze` instruction, I'm leaning towards preferring actually getting `num_indices` values out.

aszepieniec:
Thanks for the feedback!

> Why not just have this method live on AlgebraicHasher?
>
> What's the idea behind AlgebraicPrng?

The name "Hasher" does not invoke any meaningful mathematical formalism at least in my mind. What I think of when I hear the term is the `StringBuilder`-like pattern that you really only need in the context of hashing if for whatever reason the data that you are hashing does not live in contiguous memory and it's too expensive to make it contiguous before hashing. But cryptographers are allowed to make abstraction of these kind of implementation details, and therefore there is no mathematical notion that maps to this possible meaning of the term "Hasher". Hash functions do exist, but these have no state and no variable output length.

There is a well defined cryptographic notion with according security definitions related to the functionalities `sample_indices` and `sample_weights`. Specifically, these functions a) produce uniformly (pseudo)random values over some domain; b) are entirely deterministic (hence *pseudo*-random); c) update a hidden state; d) are hard to invert. The matching cryptographic primitive is a *pseudo-random number generator (PRNG)*. You can build PRNGs out of sponge constructions or out of hash functions or even block ciphers. The idea behind `AlgebraicPrng` was to capture this formalism abstractly.

That said, PRNGs and hash functions are closely related, and I'm not aware of any term that captures both under the same umbrella. I can live with "Hasher" denoting this family of closely related functions in our codebase.

> The reason sample_indices() and sample_weights() were recently removed from twenty-first is because the way weights are sampled is highly specific to Triton VM; that is, whatever way is most convenient to sample weights inside the VM should affect how we sample them in STARK/FRI (which are also in Triton VM).
> 
> I had originally placed them on this trait because I wanted to bundle everything related to hashing in one place. Ferdinand convinced me that it's probably better to have them (or at least sample_weights()) in Triton VM. (You could similarly argue that squeeze and absorb are quite coupled to Triton VM as well.)

I think that the interface should definitely live on `twenty-first`. If necessary, `triton-vm` can define its own implementation: `TritonHasher : AlgebraicHasher`. But `twenty-first` should *also* have *some* implementation of the interface, and it doesn't make any sense to have an implementation that disagrees with what Triton VM needs. So I tend to disagree with Ferdinand here: this is a case where Triton VM's architectural choices percolate upstream into `twenty-first`.

> If I understand correctly, we have two independent subscribers to sample_indices(), but only one subscriber to sample_weights(), which is Triton VM. If that is true, we might consider only having sample_indices() in twenty-first.

While that pattern of subscribers is accurate, I advocate against separating `sample_x` from `sample_y`. These are all PRNGs, invoke the same logic, and belong in the same source file.

And incidentally: *weights* is the poor name of the purpose that the sampled numbers in the perspective of the Triton prover and verifier. It is a poor name because both prover and verifier also need *indeterminates*, which are not weights but which do come from the same function. What these things are across (potential) subscribers, is pseudorandom elements from the extension field. "Pseudorandom" is obvious from the interface. And to optimize the distinction with regards to `indices` I would suggest the name `sample_scalars`.

> Calling the parameter max: u32 makes it sound like it's an inclusive upper bound, which it is not.
> 
> Calling it upper_bound: u32 leaves it ambiguous, but not indicating towards the wrong interpretation.

Agreed!

> I'm leaning towards preferring actually getting num_indices values out.

Agreed!

aszepieniec:
About the branch [asz+ssh/sample_indices](https://github.com/Neptune-Crypto/twenty-first/tree/asz%2Bssh/sample_indices): While I still want some changes, I think that branch is a better starting point than this PR.

sshine:
I'll put `sample_xfield()` back.

I completely agree that `AlgebraicPrng` is a more precise abstraction. I just wonder if we ever need to differentiate. There is some virtue in not complicating the trait hierarchy.

sshine:
> And to optimize the distinction with regards to indices I would suggest the name `sample_scalars`.

I'll add this, too.

aszepieniec:
Sorry about the confusion; let me clarify. The sampled XFieldElements, whether used as weights or as indeterminates, are all *scalars*. The sampled non-negative integers in the bounded range are only ever used as indices and might as well be named as such.

twenty-first/src/util_types/algebraic_hasher.rs
```diff
@@ -78,6 +88,50 @@ pub trait AlgebraicHasher: SpongeHasher {
...
+    ///
+    /// - The randomness depends on `state`.
+    ///
+    /// Since [RATE] is not divisible by [EXTENSION_DEGREE], produce as many [XFieldElement] per
+    /// `squeeze` as possible, and spill the remaining element(s). This causes some internal
+    /// fragmentation, but it greatly simplifies building [AlgebraicHasher::sample_xfield()] on
+    /// Triton VM.
+    fn sample_scalars(state: &mut Self::SpongeState, num_elements: usize) -> Vec<XFieldElement> {
+        // let num_squeezes = (num_elements * EXTENSION_DEGREE + Self::RATE - 1) / Self::RATE;
+        let num_squeezes = num_elements * EXTENSION_DEGREE / Self::RATE;
```
jan-ferdinand, line 122: Because of spill, I don't think this is correct. Consider:

```rust
assert_eq!(3, EXTENSION_DEGREE);
assert_eq!(10, Self::RATE);
let num_elements = 10;
let num_squeezes = num_elements * EXTENSION_DEGREE / Self::RATE; // 3, but should be 4
```

```suggestion
        const XFES_PER_SQUEEZE = Self::Rate / EXTENSION_DEGREE;
        let num_squeezes = (num_elements + XFES_PER_SQUEEZE - 1) / XFES_PER_SQUEEZE;
```

Picking up above example:

```rust
const XFES_PER_SQUEEZE = Self::Rate / EXTENSION_DEGREE; // 3
let num_squeezes = (num_elements + XFES_PER_SQUEEZE - 1) / XFES_PER_SQUEEZE; // 4
```

Consider collecting into a vector and asserting the length of the vector before returning.

twenty-first/src/util_types/algebraic_hasher.rs
```diff
@@ -205,4 +260,56 @@ mod algebraic_hasher_tests {
...
+
+    #[test]
+    fn sample_scalars_test() {
+        let amounts = [0, 1, 2, 3, 4];
+        let mut sponge = Tip5::init();
+        seed_tip5(&mut sponge);
+        let mut product = XFieldElement::one();
+        for amount in amounts {
+            let scalars = Tip5::sample_scalars(&mut sponge, amount);
+            product *= scalars.iter().fold(XFieldElement::one(), |a, f| a * *f);
```
jan-ferdinand, line 311:
```suggestion
            assert_eq!(amount, scalars.len());
            product *= scalars
                .into_iter()
                .fold(XFieldElement::one(), XFieldElement::mul);
```

jan-ferdinand (changes requested):
I think there's 2 bugs in the current `sample_scalars`. Left a few additional nits inline.

twenty-first/src/util_types/algebraic_hasher.rs
jan-ferdinand: An assertion as described can also happen in some test, but that test should be more thorough than the currently existing one.

twenty-first/src/util_types/algebraic_hasher.rs
```diff
@@ -78,6 +88,49 @@ pub trait AlgebraicHasher: SpongeHasher {
...
+        (0..num_squeezes)
+            .map(|_| Self::squeeze(state))
+            .flat_map(|elems| {
+                vec![
+                    XFieldElement::new([elems[0], elems[1], elems[2]]),
+                    XFieldElement::new([elems[3], elems[4], elems[5]]),
+                    XFieldElement::new([elems[6], elems[7], elems[8]]),
+                    // spill 1 element, elems[9], per squeeze
+                ]
+            })
```
jan-ferdinand, line 131:
```suggestion
            .flat_map(|bfe_elems| {
                let mut xfe_elems = Vec::with_capacity(xfes_per_squeeze);
                for i in 0..xfes_per_squeeze {
                    xfe_elems.push(XFieldElement::new([
                        bfe_elems[i * EXTENSION_DEGREE],
                        bfe_elems[(i * EXTENSION_DEGREE) + 1],
                        bfe_elems[(i * EXTENSION_DEGREE) + 2],
                    ]));
                }
                xfe_elems
            })
```

jan-ferdinand:
The solution in cc6c3e4 is much nicer than what I have come up with. :+1:

I wonder: is the Tip5-specific implementation in 1ce639f7 an improvement over the generic implementation? If no, is the purpose of it being there to have scalar sampling explicit, _i.e._, spelled out in detail, simplifying the implementation of the equivalent function in Triton VM?

aszepieniec:
It's not an improvement in terms of the number of invocations of the Tip5 permutation if you count the number of iterations it takes to fill an array of, say, 10 XFieldElements. But my understanding is that the spilling implementation is easier to translate into tasm. We can revise the specific implementation simultaneously with the tasm counterpart.

PR #97, opened 2023-03-04
fail on attempting to invert zero in the extension field

jan-ferdinand:


Sword-Smith:
I think we should add a "should panic" test for this case.

Sword-Smith (approved):

aszepieniec (approved):
Down the line I would prefer migrating to returning an `Option<Self>` instead.

But for now the important thing is that it agrees with `BFieldElement`'s `inverse`.

PR #98, opened 2023-03-04
make Sponge's state `Debug`

jan-ferdinand:


jan-ferdinand:
It can be pretty useful to have some access to the underlying state of the Sponge when debugging. Even though I had closed this PR previously because the immediate issue then had been resolved, a similar situation is coming up again now and might again in the future. Unless there are concerns about leaking the Sponge's state – which I'm happy to discuss but don't think are warranted by this trait bound – I think that this debugging functionality is a worthwhile addition.

Sword-Smith (approved):

PR #99, opened 2023-03-08
Simplify mmr algos

Sword-Smith:
- convert MMR indexes from `u128` to `u64`
- split shared MMR algorithms into `basic` and `advanced` where the goal is that only `basic` are part of the consensus algorithm for Neptune, and `advanced` is only needed for archival nodes.
- Make main algorithms for index arithmetic run in constant time. This applies to `right_lineage_length_from_leaf_index` which tells how many parents to add when a new leaf is inserted, and to `leaf_index_to_mt_index_and_peak_index` which converts a leaf index into the index the leaf would have if it was stored in a Merkle tree and is used in the algorithm that mutates a leaf in a MMR.
- Remove `Option` type unneeded places

The `u128` -> `u64` conversion and the removal of the `Option` type should make this code compilable with the `tasm-lang` compiler:
https://github.com/TritonVM/tasm-lang


Sword-Smith:
Before merging this, let's fix the return type of `mmr_membership_proof`'s `verify` function. And maybe also move the verify logic to `shared_basic`.

Sword-Smith:
This work is not done yet, but I'm merging so we can start indexing MMRs with `u64` instead of `u128`.

PR #101, opened 2023-03-15
simplify MMR index arithmetic

jan-ferdinand:
Thank you, OEIS: [A005187](https://oeis.org/A005187).


Sword-Smith:
If `non_leaf_nodes_left` is no longer used in `leaf_index_to_node_index`, I think `non_leaf_nodes_left` can be removed entirely.

Sword-Smith (changes requested):
If non_leaf_nodes_left is no longer used in leaf_index_to_node_index, I think non_leaf_nodes_left can be removed entirely.

jan-ferdinand:
Good catch. :+1:

Sword-Smith:
Some of the code is complicated by the `node_index` being zero-indexed instead of one-indexed. Would it be worth seeing what the effects of zero-indexing `node_index` would result in? If you want to give that a try, please do. If not, I'll have a look at it.

Sword-Smith (approved):
LGTM

PR #106, opened 2023-04-11
add canonical interpretation of XFieldElements as Digests

jan-ferdinand:
This allows more efficient Merkle tree building if the leaves don't need prior compression.


aszepieniec:
LGTM

PR #107, opened 2023-04-28
improve on Merkle tree's de-duplicated authentication structure generation & verification

jan-ferdinand:
This PR brings multiple goodies.

- A benchmark for verifying partial Merkle authentication paths.
- A tremendous speedup when verifying partial authentication paths: what used to take ~500ms now takes ~650μs. Yes, you've read that right, the new code for verifying the particular de-duplicated Merkle tree authentication structure from the new benchmark shaves off ~99.87%[^2] of the runtime. This is achieved by performing as little hashing as possible.[^1]
- The expected height of the Merkle tree for the which the authentication structure is claimed to be valid is now an obligatory parameter. If any partial authentication path does not conform to this expected height, validation will fail.
- A bug fix in the logic of generating partial authentication paths.
- Big code readability improvements for both generation and verification of partial authentication paths.
- Implementations of traits `Deref` and `DerefMut` for partial authentication paths. This allows easier access to the underlying vector. For example, `auth_path.0.len()` can now be `auth_path.len()`.

[^1]: In contrast to the old logic, the new code for verifying partial authentication paths is completely sequential. It is possible that additional speedups can be achieved by introducing parallelization.
[^2]: On the beefy machine known as “mjolnir”, the respective times are 170ms and ~80μs, translating to 99.95%.

### DIFF
--- a/twenty-first/Cargo.toml
+++ b/twenty-first/Cargo.toml
@@ -101,3 +101,7 @@ harness = false
 [[bench]]
 name = "merkle_tree"
 harness = false
+
+[[bench]]
+name = "merkle_tree_authenticate"
+harness = false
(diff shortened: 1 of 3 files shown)


Sword-Smith:
Fix arithmetic that could overflow, in 8869e4484e767b3972b4c49a6cd846286e901c62. Otherwise very nice!!!

Sword-Smith (approved):

PR #112, opened 2023-05-04
add new function for digest: vmhash

aszepieniec:


jan-ferdinand:
No objections, just questions:
1. In which scenario(s) is it useful (necessary?) to hash a digest, given that both input and output to the new function are digests, with nothing else mixed in?
2. Minor: the twenty-first repository tries to be somewhat independent of the Triton VM repository. Of course, the influence is often circular, else there wouldn't be the need for new hash functions. While this method is not the first explicit reference to Triton VM[^1], it might be worth considering keeping the circular references small. That said, hunting for a better name just for the sake of a better name is not worth it should such a hunt prove difficult.

[^1]: The only other reference I can find is in the docstring for `fn sample_scalars` in the `impl AlgebraicHasher for Tip5`.

jan-ferdinand (approved):
While it does strike me as a little odd to hash only a single digest, I don't have enough domain-specific knowledge to suggest alternatives and judge how reasonable they are.

jan-ferdinand:
With TritonVM/triton-vm#194 merged, Triton VM only relies on the trait `Hashable` for hashing `Program`s. As far as I can tell, this does not interfere with removing the implementation of `Hashable` for `Digest`.

PR #119, opened 2023-05-23
Add derive macro for BFieldCodec

Sword-Smith:
Currently only works for structs with named fields. Expand.


Sword-Smith:
We should add functionality for deriving `BFieldCodec` for structs with unnamed fields (tuple structs) and for enums. Potentially also for structs with fields of type `Vec<T>` where `T` implements `BFieldCodec`.

Sword-Smith:
Still missing derive rules for `Vec<T>`

Sword-Smith:
We should verify that we support type parameters and `PhantomData` in the structs before merging this.

Sword-Smith:
I think this is in a good state now, ready to be merged.

Sword-Smith:
@jan-ferdinand This might cause trouble for Triton-VM. Let's asses when you have time.

PR #120, opened 2023-06-09
make `BFieldCodec` derive macro more stable and feature complete

jan-ferdinand:
- Use the new macro in more places.
- Add documentation, including example and known limitations.
- Allow ignoring fields via attribute `#[bfield_codec(ignore)]`. Ignored fields must implement `Default`.
- Make usage of the macro more stable through fully qualified names.
- Slight code cleanup.

### DIFF
--- a/twenty-first/Cargo.toml
+++ b/twenty-first/Cargo.toml
@@ -29,7 +29,8 @@ features = ["precommit-hook", "run-cargo-clippy", "run-cargo-fmt"]
 [dependencies]
 anyhow = "1.0"
 bincode = "1.3"
-bfieldcodec_derive = "0.2.0"
+# bfieldcodec_derive = "0.2.0"
+bfieldcodec_derive = { path = "../bfieldcodec_derive" }
 blake3 = "1.3.3"
 byteorder = "1.4"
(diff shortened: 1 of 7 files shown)


twenty-first/Cargo.toml
```diff
@@ -29,7 +29,8 @@ features = ["precommit-hook", "run-cargo-clippy", "run-cargo-fmt"]
 [dependencies]
 anyhow = "1.0"
 bincode = "1.3"
-bfieldcodec_derive = "0.2.0"
+# bfieldcodec_derive = "0.2.0"
+bfieldcodec_derive = { path = "../bfieldcodec_derive" }
```
aszepieniec, line 33: This does not seem like it belongs in a commit, does it?

aszepieniec (approved):
Great job!

twenty-first/Cargo.toml
jan-ferdinand: Good point. We should release a new version of `bfieldcodec_derive` and depend on that in `twenty-first`, and potentially release a new version of `twenty-first` after that.

PR #125, opened 2023-06-22
improve derivation of `BFieldCodec` for structs with only fixed-length fields

### LINKED ISSUE #124: Error in derived implementation of `BFieldCodec` for vectors
Minimal example

```rust
use twenty_first::shared_math::tip5::Digest;

#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, GetSize, BFieldCodec)]
pub struct Foo {
    pub bar: Digest,
}

#[cfg(test)]
mod tests {
    use rand::random;
    use super::*;

    #[test]
    fn foo_vec_test() {
        for i in 0..5 {
            let foos = vec![
                Foo {
                    bar: random(),
                };
                i
            ];
            let encoded = foos.encode();
            let decoded = *Vec::<Foo>::decode(&encoded).unwrap();
            assert_eq!(foos, decoded);
        }
    }

```

Error message:
```
thread 'foo_vec_test' panicked at 'called `Result::unwrap()` on an `Err` value: Length indication plus one must match actual sequence length. Claimed vector length was 1. Item size was 5. Sequence length was 7.', ...
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The encoding prepends the size of each element, but the decoder does not expect that.

jan-ferdinand:
Fix #124

### DIFF
--- a/twenty-first/src/shared_math/bfield_codec.rs
+++ b/twenty-first/src/shared_math/bfield_codec.rs
@@ -461,7 +461,9 @@ mod tests {
                 assert!(
                     decoded_mutated.is_err()
                         || *decoded_mutated.unwrap() != original_decoded
-                        || mutated_encoding[i] == original_value
+                        || mutated_encoding[i] == original_value,
+                    "Error: Mutated encoding was: {}",
+                    mutated_encoding.iter().join(",")
                 );
             }
 
@@ -473,7 +475,9 @@ mod tests {
                 assert!(
                     decoded_mutated.is_err()
                         || *decoded_mutated.unwrap() != original_decoded
-                        || mutated_encoding[i] == original_value
+                        || mutated_encoding[i] == original_value,
+                    "Error: Mutated encoding was: {}",
+                    mutated_encoding.iter().join(",")
                 );
             }
 
--- a/twenty-first/src/shared_math/bfield_codec.rs
+++ b/twenty-first/src/shared_math/bfield_codec.rs
@@ -10,11 +10,8 @@ use num_traits::Zero;
 // an explicit dependency on `bfieldcodec_derive` to their Cargo.toml.
 pub use bfieldcodec_derive::BFieldCodec;
 
-use crate::util_types::algebraic_hasher::AlgebraicHasher;
-
 use super::b_field_element::BFieldElement;
 use super::tip5::Digest;
-use super::tip5::Tip5;
 use super::tip5::DIGEST_LENGTH;
 use super::x_field_element::XFieldElement;
 use super::x_field_element::EXTENSION_DEGREE;
@@ -229,6 +226,10 @@ impl<T: BFieldCodec, S: BFieldCodec> BFieldCodec for (T, S) {
                 length
             }
         };
+
+        if str.len() < index + len_t {
+            bail!("Sequence too short to decode tuple, element 0");
+        }
         let t = *T::decode(&str[index..index + len_t])?;
         index += len_t;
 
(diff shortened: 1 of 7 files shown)


twenty-first/src/shared_math/bfield_codec.rs
```diff
@@ -707,7 +699,11 @@ mod bfield_codec_tests {
     fn test_decode_random_negative() {
         for _ in 1..=10000 {
             let len = random_length(100);
-            let str: Vec<BFieldElement> = random_elements(len);
+            let mut str: Vec<BFieldElement> = random_elements(len);
+            // Claiming a length that is too large leads to error “capacity overflow” when decoding.
+            if !str.is_empty() {
+                str[0] = (100 + random_length(500) as u64).into();
+            }
```
jan-ferdinand, line 706: I'm not super happy about this. The underlying requirement comes from initiating the result vector as `Vec::with_capacity(num_elements)` in `impl<T> BFieldCodec for Vec<T>::decode()`, which is nice to have but apparently introduces a DOS vector. The alternative is to use `vec![]` and bite the cost of internal re-allocations.

Thought, @Sword-Smith?

jan-ferdinand: After writing above comment, I feel much more strongly inclined to use `vec![]` instead of `Vec::with_capacity(random_persons_claim)`.

Sword-Smith (changes requested):
This certainly works in isolation, and it might be an improvement.

A significant change, though, is that the encoding of a `struct` no longer prepends the length of each field. This will break things downstream, especially things in https://github.com/TritonVM/tasm-lib/tree/asz/removal-record-integrity/src/structure.

It might be advantageous to always prepend the field length when encoding a struct, even if it is known at compile-time as that makes any TASM program that deals with these encodings simpler.

I think @aszepieniec should weigh in on this.

This new rule that @jan-ferdinand has programmed would mean that the field size is *not* prepended if it is known statically, otherwise the field size *is* prepended. It makes for shorter encodings, but also for potentially more complicated TASM programs to get a specific field from a struct.

jan-ferdinand:
Good point. I wasn't aware of such dependencies downstream. In isolation, it seems to me that my suggested solution is a rather clean approach. My hope is that the approach can be transferred to the TASM context as well.

aszepieniec:
If every field is length-prepended, then the procedure for "get field x" can be struct-agnostic simply by indexing and hopping over fields until a decrementing counter reaches zero. But if some fields are not length-prepended, then the "get field x" method has to be a custom method for every struct, for every field.

I started implementing a collection of "get field x" methods for `TransactionKernel` assuming (wrongly) that statically-sized fields are not length-prepended. The result was a monster. I'm very happy to use the generic "get nth field" method that works for any struct under the length-prepended encoding and I would strongly suggest we not change this property.

Sword-Smith:
After talking to @aszepieniec we have agreed that this proposal from @jan-ferdinand is the way we want to go. But please don't merge before conferring with @aszepieniec or me, as there will be downstream complications.

Sword-Smith:
@aszepieniec and myself should go over this new `BFieldCodec` derivation before merging, since we have the most experience working with these encodings in TASM.

- [ ] Approvable by @aszepieniec 
- [ ] Approvable by @Sword-Smith 
- [ ] Approvable by @jan-ferdinand (optional)

Sword-Smith (changes requested):
No merge before approved by requested reviewers

aszepieniec:
Approved via #136

PR #128, opened 2023-06-28
simpler Merkle tree authentication structure

### LINKED ISSUE #123: reduce size of `BFieldCodec`-encoded authentication structure
An “authentication structure” removes redundancies in a collection of Merkle authentication paths. Two authentication paths for the same Merkle tree are bound to overlap. For example, in the tree below, the authentication path for the leaf marked by the blue arrow is indicated by blue nodes, and similarly for the leaf marked by the pink arrow.

image

When verifying the blue authentication path, the node encircled in pink is computed. Therefore, it does not have to be supplied if _both_ paths are being verified. The same holds true for the node encircled in blue.

In our current implementation, an authentication structure is a `Vec<Vec<Option<Digest>>>`. Omitted nodes are represented by a `None`, included nodes are `Some(digest)`. Each authentication path is granted its own `Vec<_>`. However, the indices of the leaves whose inclusion in the tree is to be proven identify all nodes that can be omitted. An authentication structure could well be a `Vec<Digest>`.

Some quick experiments show that for a Merkle tree with $2^{15}$ leaves, an authentication structure for 32 revealed elements requires a total of 2487 base field elements for encoding. Of those, 388 (15.6%) are required for encoding `None`s. Another 32 (1.3%) are used for length-prepending the inner vectors. In total, the encoded size of an authentication structure could be reduced by about 15%[^1][^2].

[^1]: Above numbers come from a single measurement. Different leaf indices and tree sizes will give different results.
[^2]: In Triton VM, the total reduction of the proof size is estimated at around 3%.

jan-ferdinand:
- more compact authentication structure
- cleaner generation of authentication structure

Fix #123.

### DIFF
--- a/twenty-first/Cargo.toml
+++ b/twenty-first/Cargo.toml
@@ -84,3 +84,7 @@ harness = false
 [[bench]]
 name = "merkle_tree_authenticate"
 harness = false
+
+[[bench]]
+name = "merkle_tree_auth_structure_size"
+harness = false
--- a/twenty-first/benches/merkle_tree_auth_structure_size.rs
+++ b/twenty-first/benches/merkle_tree_auth_structure_size.rs
@@ -1,0 +1,117 @@
+use criterion::criterion_group;
+use criterion::criterion_main;
+use criterion::measurement::Measurement;
+use criterion::measurement::ValueFormatter;
+use criterion::BenchmarkId;
+use criterion::Criterion;
+use criterion::Throughput;
+use itertools::Itertools;
+use rand::rngs::StdRng;
+use rand::Rng;
+use rand::RngCore;
+use rand::SeedableRng;
+use twenty_first::shared_math::bfield_codec::BFieldCodec;
+
+use twenty_first::shared_math::tip5::Tip5;
+use twenty_first::util_types::algebraic_hasher::AlgebraicHasher;
+use twenty_first::util_types::merkle_tree::CpuParallel;
+use twenty_first::util_types::merkle_tree::MerkleTree;
+use twenty_first::util_types::merkle_tree_maker::MerkleTreeMaker;
+
+#[derive(Debug, Clone, Copy)]
+struct AuthStructureEncodingLength(f64);
+
+#[derive(Debug, Clone, Copy)]
+struct AuthStructureEncodingLengthFormatter;
+
+impl Measurement for AuthStructureEncodingLength {
+    type Intermediate = ();
+    type Value = Self;
+
+    fn start(&self) -> Self::Intermediate {
+        ()
+    }
+
+    fn end(&self, _i: Self::Intermediate) -> Self::Value {
+        self.to_owned()
+    }
+
+    fn add(&self, v1: &Self::Value, v2: &Self::Value) -> Self::Value {
+        AuthStructureEncodingLength(v1.0 + v2.0)
+    }
+
+    fn zero(&self) -> Self::Value {
+        AuthStructureEncodingLength(0.0)
+    }
+
+    fn to_f64(&self, value: &Self::Value) -> f64 {
+        value.0
+    }
+
+    fn formatter(&self) -> &dyn ValueFormatter {
+        &AuthStructureEncodingLengthFormatter
+    }
+}
+
+impl ValueFormatter for AuthStructureEncodingLengthFormatter {
+    fn scale_values(&self, _typical_value: f64, _values: &mut [f64]) -> &'static str {
+        "bfe"
+    }
+
+    fn scale_throughputs(
+        &self,
+        _typical_value: f64,
+        _throughput: &Throughput,
+        _values: &mut [f64],
+    ) -> &'static str {
+        "bfe/s"
+    }
+
+    fn scale_for_machines(&self, _values: &mut [f64]) -> &'static str {
+        "bfe"
+    }
+}
+
+fn auth_structure_len(c: &mut Criterion<AuthStructureEncodingLength>) {
+    let mut rng = StdRng::seed_from_u64(0);
+
+    let tree_height = 22;
+    let num_leaves = 1 << tree_height;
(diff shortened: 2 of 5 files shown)


jan-ferdinand:
More extensive analysis of the data that is now available indicates that the expected improvement of the size of the encoding of the authentication structure is around 30%.

For example, the average encoded size of the current authentication structure for 40 revealed indices can be modeled as $f = 236\cdot\texttt{tree\\_height} + 634$. The average encoded size of the new authentication structure for 40 revealed indices can be modeled as $g = 160\cdot\texttt{tree\\_height} + 147$. The limit of the ratio of the improvement, _i.e._, of $\frac{g - f}{f}$, approaches ~32.2% from above. In other words, the expected size improvement is _at least_ 32% for 40 revealed indices.

The same holds true for different number of revealed indices: the expected size improvement is at least 28.5% for 1 revealed index, at least 30.7% for 2 revealed indices, 31.5% for 3 revealed indices, and so on.

jan-ferdinand:
The data for above claims was generated using the following procedure.

```rust
fn print_merkle_tree_authentication_structure_sizes() {
    type H = Tip5;
    type M = CpuParallel;
    type MT = MerkleTree<H>;

    let tree_heights = 2..10;
    let num_samples = 100;
    let mut all_sizes = vec![];
    for tree_height in tree_heights.clone() {
        let num_leaves = 1 << tree_height;
        let leaves = random_elements(num_leaves);
        let tree: MT = M::from_digests(&leaves);
        let mut this_tree_sizes = vec![0; num_leaves];
        for num_revealed_indices in 0..num_leaves {
            for _ in 0..num_samples {
                let revealed_indices =
                    random_elements_distinct_range(num_revealed_indices, 0..num_leaves);
                let auth_structure = tree.get_authentication_structure(&revealed_indices);
                let encoding = auth_structure.encode();
                this_tree_sizes[num_revealed_indices] += encoding.len();
            }
        }
        all_sizes.push(this_tree_sizes);
    }

    println!();
    println!("| tree_height | average size of auth structure per num_revealed_leaves |");
    println!("|---:|:---|");
    for (this_tree_sizes, tree_height) in all_sizes.iter().zip(tree_heights) {
        let this_tree_sizes_avg = this_tree_sizes
            .iter()
            .map(|total| total / num_samples)
            .collect_vec();
        let this_tree_sizes_str = this_tree_sizes_avg.iter().join(", ");
        println!("| {tree_height} | {this_tree_sizes_str} | ");
    }
}
```

Sword-Smith (changes requested):
- `num_leaves` is technically the number of leaves plus one. Can we find a better name?
- The verifier seems to use a lot of RAM for `partial_merkle_tree`. Can't we use a `HashMap` here instead?

jan-ferdinand:
> * `num_nodes` is technically the number of nodes plus one. Can we find a better name?

I would argue that the name is in line with using `self.nodes` to refer to all nodes plus the one dummy element at index 0.

>     * The verifier seems to use a lot of RAM for `partial_merkle_tree`. Can't we use a `HashMap` here instead?

Good point. I will investigate what that looks like.

>     * Why are the fields of `MerkleTree` public? Do they have to be?

They do not. I suppose now is as good a time as any to change that. :+1:

jan-ferdinand:
> They do not. I suppose now is as good a time as any to change that. +1

If we're in general Merkle tree cleanup mode: do we want to get rid of methods that are only used in tests in this same PR? Examples are `root_from_arbitrary_number_of_digests`, `get_authentication_path` (interestingly), and `verify_authentication_path_from_leaf_hash`. As far as I can see none of our projects seem to use those methods.

jan-ferdinand:
> > The verifier seems to use a lot of RAM for `partial_merkle_tree`. Can't we use a `HashMap` here instead?
> 
> Good point. I will investigate what that looks like.

The change incurs a small slowdown[^0]. The improved memory consumption probably makes that worthwhile.

[^0]: on my machine, ~27ms vs ~32ms for the benchmark `merkle_tree_authenticate`

jan-ferdinand:
> Aren't `get_authentication_path` and `verify_authentication_path_from_leaf_hash` the _basic_ functions for a Merkle tree? It seems weird to me not to have those functions.

I concur with the feeling. As you have pointed our, calling the `[…]_authentication_structure` methods with just a single index is equivalent to the `[…]_authentication_path` methods. I have updated the documentation accordingly and removed the basically duplicate methods, as discussed irl.

> I don't know how many thoughts you've done about the recufier.

Admittedly, not many. The slight, but as of yet unfounded, hope is that the new methods for authentication structure verification will translate nicely to Triton assembly, too. In the worst case, we can fall back to naïve verification.

Sword-Smith:
I get a *very* significant speedup if I change the Merkle Tree authentication benchmark parameters used in Triton VM.

```rust
-    let tree_height = 16;
+    let tree_height = 22;
     let num_leaves = 1 << tree_height;
     let leaves = (0..num_leaves).map(|_| rng.next_u64()).collect_vec();
     let leaf_digests = leaves.iter().map(Tip5::hash).collect_vec();
     let mt: MerkleTree<Tip5> = CpuParallel::from_digests(&leaf_digests);
     let mt_root = mt.get_root();
 
-    let num_opened_indices = num_leaves / 4;
+    let num_opened_indices = 40;
     let opened_indices = (0..num_opened_indices)
         .map(|_| rng.gen_range(0..num_leaves))
         .collect_vec();
```

82ms with `Vec<Option<Digest>>` and 3.43ms with `HashMap`. A 24x speedup!

jan-ferdinand:
With the latest changes in place, not only is the generated authentication structure considerably smaller, the performance of both generation and verification of a Merkle tree's authentication structure has improved dramatically for the more realistic parameters. On my machine:

- Generation used to take ~580μs, now takes ~75μs (reduction by ~87%).
- Verification used to take ~100ms, now takes ~600μs (reduction by ~99%) (!).

Sword-Smith (approved):

PR #137, opened 2023-07-30
Make tip5 permutation public

igorsyl:


aszepieniec:
Hi Igor, thanks for the contribution! Can you comment on why the suggested change is an improvement? From our perspective, you should only be allowed to invoke the permutation if you're using the sponge in a particular mode, for instance hashing fixed length inputs or variable length inputs.

igorsyl:
Thank you for your quick reply! I’m adding your cool new hash function to Plonky3 and they expose the permutation. I’m still looking as to why permutation is needed in that repo and not just the digest hash. 

https://github.com/Plonky3/Plonky3/pull/42

dlubarov:
Yeah, in Plonky3 we ideally like to have access to underlying cryptographic permutations, so we can use them in other constructions like

- variants of sponges (different padding rules, overwrite mode, etc)
- specialized compression functions, like the ones in the Anemoi or Monolith papers
- truncated permutations (as used e.g. [here](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=9d0d40b17a6f754836cbca363a6587d1efc01511))

Would you have any concerns with the Tip5 permutation being used as a generic cryptographic permutation, in different constructions where certain security properties have been shown under a random permutation assumption?

PR #140, opened 2023-08-31
Add BFieldCodec support for structs containing empty structs

Sword-Smith:
There was a cornercase here that was not covered: A struct containing a field of a type that encodes to an empty list. Also fixes the equivalent problem for `enum` containing an empty struct as associated data to a variant.

Note that the `bfieldcodec_derive` import was changed to a local version since the added test would otherwise fail.

### DIFF
--- a/bfieldcodec_derive/src/lib.rs
+++ b/bfieldcodec_derive/src/lib.rs
@@ -479,7 +479,8 @@ fn generate_decode_statement_for_field(
     let field_name_as_string_literal = field_name.to_string();
     quote! {
         let (#field_name, sequence) = {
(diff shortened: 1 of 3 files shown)


Sword-Smith:
The `derive` macro currently fails when attempting to derive `BFieldCodec` for an enum with no possible values, like so: `enum Impossible {}`. But that is a value that can never be created, so perhaps we don't care.

cf. https://stackoverflow.com/questions/47589690/what-does-a-match-on-an-empty-enum-return

jan-ferdinand:
Fixed by 164bdcf1a29180364945932862b09ea4a04ac1c3

PR #141, opened 2023-09-18
allow (i)NTT on empty input

jan-ferdinand:


Sword-Smith (approved):
Nice. Good test. Would it be possible to handle the empty case without a new branch and return statement? Could be cool if the same codepath was always followed. If that can't be done easily, please merge.

jan-ferdinand:
The code in the `ntt` method is so difficult to digest that I don't currently want to spend the time figuring out whether that is possible.

PR #142, opened 2023-09-19
use property based tests for BFieldElement

jan-ferdinand:
Refactor and clean up some BFieldElement tests so that they only do one thing. Use an actual property testing framework for property based tests.

### DIFF
--- a/twenty-first/src/shared_math/b_field_element.rs
+++ b/twenty-first/src/shared_math/b_field_element.rs
@@ -689,10 +689,27 @@ mod b_prime_field_element_test {
         assert!(zero.is_zero());
     }
 
+    proptest! {
+        #[test]
+        fn not_zero_is_nonzero(bfe in arbitrary_non_zero_bfield_element()) {
+            prop_assert!(!bfe.is_zero());
+        }
+    }
+
     #[test]
     fn one_is_one() {
         let one = BFieldElement::one();
         assert!(one.is_one());
+    }
+
+    proptest! {
+        #[test]
+        fn not_one_is_not_one(bfe: BFieldElement) {
+            if bfe.value() == 1 {
+                return Ok(());
+            }
+            prop_assert!(!bfe.is_one());
+        }
     }
 
     #[test]
--- a/twenty-first/src/shared_math/b_field_element.rs
+++ b/twenty-first/src/shared_math/b_field_element.rs
@@ -552,22 +552,129 @@ mod b_prime_field_element_test {
     use crate::shared_math::polynomial::Polynomial;
     use itertools::izip;
     use proptest::prelude::*;
-    use rand::{random, thread_rng};
+    use rand::thread_rng;
 
-    #[test]
-    fn get_size_test() {
-        let bfe: BFieldElement = random();
-        assert_eq!(8, bfe.get_size());
+    impl Arbitrary for BFieldElement {
+        type Parameters = ();
+        fn arbitrary_with(_: Self::Parameters) -> Self::Strategy {
+            (0..BFieldElement::P).prop_map(BFieldElement::new).boxed()
+        }
+
+        type Strategy = BoxedStrategy<Self>;
+    }
+
+    prop_compose! {
+        fn arbitrary_non_zero_bfield_element()(x in 1..BFieldElement::P) -> BFieldElement {
+            BFieldElement::new(x)
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn get_size(bfe: BFieldElement) {
+            prop_assert_eq!(8, bfe.get_size());
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn zero_is_neutral_element_for_addition(bfe: BFieldElement) {
+            let zero = BFieldElement::zero();
+            prop_assert_eq!(bfe + zero, bfe);
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn one_is_neutral_element_for_multiplication(bfe: BFieldElement) {
+            let one = BFieldElement::one();
+            prop_assert_eq!(bfe * one, bfe);
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn addition_is_commutative(element_0: BFieldElement, element_1: BFieldElement) {
+            prop_assert_eq!(element_0 + element_1, element_1 + element_0);
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn multiplication_is_commutative(element_0: BFieldElement, element_1: BFieldElement) {
+            prop_assert_eq!(element_0 * element_1, element_1 * element_0);
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn addition_is_associative(
+            element_0: BFieldElement,
+            element_1: BFieldElement,
+            element_2: BFieldElement
+        ) {
+            prop_assert_eq!(
+                (element_0 + element_1) + element_2,
+                element_0 + (element_1 + element_2)
+            );
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn multiplication_is_associative(
+            element_0: BFieldElement,
+            element_1: BFieldElement,
+            element_2: BFieldElement
+        ) {
+            prop_assert_eq!(
+                (element_0 * element_1) * element_2,
+                element_0 * (element_1 * element_2)
+            );
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn multiplication_distributes_over_addition(
+            element_0: BFieldElement,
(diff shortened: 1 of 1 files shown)


twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -552,286 +552,251 @@ mod b_prime_field_element_test {
...
 
-    #[test]
-    fn simple_value_test() {
-        let zero: BFieldElement = BFieldElement::new(0);
-        assert_eq!(0, zero.value());
-        let one: BFieldElement = BFieldElement::one();
-        assert_eq!(1, one.value());
+    proptest! {
+        #[test]
+        fn multiplication_is_associative(
```
Sword-Smith, line 625: Nice mathy tests!

twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -552,286 +552,251 @@ mod b_prime_field_element_test {
...
+            prop_assert!((bfe / bfe).is_one());
+        }
+    }
+
+    proptest! {
+        #[test]
+        fn values_larger_than_modulus_are_handled_correctly(large_value in BFieldElement::P..) {
+            let bfe = BFieldElement::new(large_value);
+            let expected_value = large_value - BFieldElement::P;
+            prop_assert_eq!(expected_value, bfe.value());
```
Sword-Smith, line 670: Wouldn't this test be marginally more readable as a one-liner? 

```rust
prop_assert_eq!(large_value - BFieldElement::P, BFieldElement::new(large_value).value());
```

twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -552,286 +552,251 @@ mod b_prime_field_element_test {
...
         }
+    }
 
-        let empty_inversion = BFieldElement::batch_inversion(vec![]);
-        assert!(empty_inversion.is_empty());
+    proptest! {
+        #[test]
+        fn value_is_preserved(value in 0..BFieldElement::P) {
+            let bfe = BFieldElement::new(value);
+            prop_assert_eq!(value, bfe.value());
```
Sword-Smith, line 741: This is also marginally more readable as a one-liner, I think.

twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -552,286 +552,251 @@ mod b_prime_field_element_test {
...
     }
 
-    #[test]
-    fn inversion_property_based_test() {
-        let elements: Vec<BFieldElement> = random_elements(30);
+    proptest! {
+        #[test]
+        fn lift_then_unlift_preserves_element(bfe: BFieldElement) {
+            let maybe_same_bfe = bfe.lift().unlift();
+            prop_assert_eq!(Some(bfe), maybe_same_bfe);
```
Sword-Smith, line 760: one-liner?

twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -552,286 +552,251 @@ mod b_prime_field_element_test {
...
+        assert_eq!("-15", format!("{minus_fifteen}"));
     }
 
     #[test]
-    fn lift_property_test() {
-        let elements: Vec<BFieldElement> = random_elements(100);
-        for element in elements {
-            assert_eq!(Some(element), element.lift().unlift());
-        }
+    fn zero_is_zero() {
```
Sword-Smith, line 687: `zero_is_zero` and `one_is_one` should also check that `one.is_zero()` and `zero.is_one()` returns false, and that an arbitrary value (can be hardcoded) returns false on `is_zero()` and `is_one()`.

twenty-first/src/shared_math/b_field_element.rs
```diff
@@ -552,286 +552,251 @@ mod b_prime_field_element_test {
...
+        let zero = BFieldElement::zero();
+        assert!(zero.is_zero());
     }
 
     #[test]
-    fn next_and_previous_test() {
```
Sword-Smith, line 674: This test checked that `increment` and `decrement` handled the corner-cases correctly wrapping around the prime. I think your property-based tests don't hit those corner-cases.

Sword-Smith (changes requested):

Sword-Smith (changes requested):
Running `cargo t` produced a file that is not committed to git. What should we do with this file, and does it somehow indicate an error? All tests have always passed as I ran them on my machine.

```
$ cat twenty-first/proptest-regressions/shared_math/b_field_element.txt 
# Seeds for failure cases proptest has generated in the past. It is
# automatically read and these particular cases re-run before any
# novel cases are generated.
#
# It is recommended to check this file in to source control so that
# everyone who runs the test benefits from these saved cases.
cc 475795502dfa9fb101aaf413911719f066f2e8698d634e9b11fc594ef3d8f71d # shrinks to byte_array = [1, 0, 0, 0, 255, 255, 255, 255]
```

twenty-first/src/shared_math/b_field_element.rs
jan-ferdinand: I've added the tests `not_zero_is_nonzero` and `not_one_is_not_one` that cover these properties more fundamentally, and as their own tests, not muddled in with an existing test. Also note the existence of test `one_unequal_zero`.

twenty-first/src/shared_math/b_field_element.rs
jan-ferdinand: I disagree with this one: the suggested one-liner is (a) more difficult to digest and (b) exceeds the line length limit.

twenty-first/src/shared_math/b_field_element.rs
jan-ferdinand: You are right, they corners are hit too rarely. I've added the tests `increment_overflow` and `decrement_underflow` for the two corner cases.

jan-ferdinand:
> Does [the file] somehow indicate an error?

This seems to be a [known bug](https://github.com/proptest-rs/proptest/issues/263) of proptest for tests marked as `should_panic`. Therefore, it does not indicate an error in this specific case, but it is also quite misleading.

> What should we do with this file?

Good question. I'm not sure. I think committing it once an actual violated property has been identified makes good sense. If my understanding of how proptest works is correct, committing it now also does not hurt, but my understanding of proptest's internals is only starting to develop.

twenty-first/src/shared_math/b_field_element.rs
Sword-Smith: Great. I don't like the names, though. Maybe `increment_wrap`, `decrement_wrap` instead? Underflow refers to a floating point becoming zero, not to a wrap around from zero to the highest possible value.

Sword-Smith (approved):

Sword-Smith (changes requested):
The generation of the proptest error file seems like a regression to me. We could consider ignoring it in `.gitignore`. That way is doesn't pollute your git interface/command line output with problematic uncommitted files.

Sword-Smith:
Another solution would be to let failing tests be regular tests, and not proptests. Concretely that would mean that `byte_array_outside_range_cannot_be_converted` shouldn't be a proptest.

twenty-first/src/shared_math/b_field_element.rs
jan-ferdinand: I've made it even more explicit with `incrementing_max_value_wraps_around` and `decrementing_min_value_wraps_around`.

jan-ferdinand:
The correct solution is to not implement `From<[u8; 8]>` but `TryFrom<[u8; 8]>`. The documentation of the `From` trait literally says:

> **This trait must not fail.** The `From` trait is intended for perfect conversions. If the conversion can fail or is not perfect, use `TryFrom`.

In this case, the test doesn't have to panic, but the returned `Result<_, _>` can be checked for `is_err()`.

Sword-Smith:
> The correct solution is to not implement `From<[u8; 8]>` but `TryFrom<[u8; 8]>`. The documentation of the `From` trait literally says:
> 
> > **This trait must not fail.** The `From` trait is intended for perfect conversions. If the conversion can fail or is not perfect, use `TryFrom`.
> 
> In this case, the test doesn't have to panic, but the returned `Result<_, _>` can be checked for `is_err()`.

Good point. I have no problem with replacing the `From<[u8; 8]>` implementation with a `TryFrom` implementation. Perhaps we should get the thumbs up from @aszepieniec before doing that though, since, AFAIK, he introduced this trait implementation when implementing Tip5.

jan-ferdinand:
With #146 merged, no spurious proptest failure file is being generated, resolving (as far as I can see) the last remaining change request.

PR #144, opened 2023-09-20
Avoid new branch in NTT

Sword-Smith:
We can avoid an extra branch in this code by requiring the provided log2 value to be zero for an empty array.


Sword-Smith:
Let me know if you think this makes sense, @jan-ferdinand.

Sword-Smith:
The force-pushed commit avoids complications of the `log_2_n_value`, I think.

jan-ferdinand:
I don't think the intention of the checks is easy to understand. That said, it seems to fit the current style of the method. :shrug:

jan-ferdinand (approved):

PR #145, opened 2023-09-22
fix edge case bugs in (i)NTT

jan-ferdinand:
The new test in this PR depends on the implementation of the `Arbitrary` trait for `BFieldElement` that is introduced in #142. The bug fix does not.


Sword-Smith (approved):
Looks good. You don't want an explicit test for input of length one? 

Sorry for introducing this bug. Thank you for fixing it.

jan-ferdinand:
I don't think you introduced it. :relieved: (At least not recently. :wink:) Happy to add the explicit test. :+1:

PR #147, opened 2023-09-25
avoid borrowing copy-type `Digest`

jan-ferdinand:
Most of the changes in this PR are boring deletion or addition of `&`s. The following things might be of interest:
- changes to the interface `hash_pair` in `AlgebraicHasher`
- commit 05a0da6b4758e11cc839dc80eb3356655663467e, introducing a benchmark
- commit e38a966acd24aa3b2b94e1620e0df94adca20da9, avoiding some copying

### DIFF
--- a/twenty-first/benches/tip5.rs
+++ b/twenty-first/benches/tip5.rs
@@ -1,6 +1,5 @@
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
-use itertools::Itertools;
-use rand::RngCore;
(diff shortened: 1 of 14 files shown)


Sword-Smith:
Looks good to me. I saw no difference in running times with this change.

Sword-Smith (approved):
Looks good to me. This change will bring the `hash_pair` signature closer to that used in the compiler or in `tasm-lib`.

I saw no difference in running times with this change.

PR #149, opened 2023-10-02
upgrade dependencies

jan-ferdinand:


Sword-Smith (approved):
Looks good! I think that this is a major version because of the dependency upgrades though. Just FYI.

PR #150, opened 2023-10-02
implement `Arbitrary` for various `struct`s

jan-ferdinand:


Sword-Smith:
Watch out for runtimes of tests. With this commit the test `shared_math::ntt::fast_ntt_attempt_tests::ntt_then_intt_is_identity_operation` goes from taking 17 seconds to taking 125 seconds. The 2nd longest test currently takes 8 seconds on my machine. You can use this syntax to reduce the number of test cases. Default is 256, so maybe 10 is appropriate?

```rust
proptest! {
        #![proptest_config(ProptestConfig {
            cases: 10, .. ProptestConfig::default()
          })]
        #[test]
        fn ntt_then_intt_is_identity_operation( ...
```

jan-ferdinand:
Interesting regression. Thanks for pointing it out. :+1:

PR #154, opened 2023-11-09
fix: extended euclidean algorithm does not panic on input 0

jan-ferdinand:


Sword-Smith:
I very much like the thorough property-based mathy tests you've added in `polynomial.rs`! Nice!

One question, though: Why do you want to change this behavior of xgcd?

jan-ferdinand:
> Why do you want to change this behavior of xgcd?

Because the current behavior is wrong: the gcd of 0 and 0 is well defined – it is 0 – and should not cause a panic.

PR #158, opened 2023-11-13
use property test framework more extensively, fix minor bug

jan-ferdinand:
- use crate `test_strategy` to increase readability of property tests
- use `proptest` framework for all ad-hoc property tests in `polynomial.rs`
- remove a substantial amount of duplicate tests
- remove duplicate method `new_const`, use `from_constant` instead
- fix bug in `fast_zerofier` returning zero instead of one on empty input
- improve panic messages and documentation of `fast_coset_evaluate` & `fast_coset_divide`


twenty-first/src/shared_math/polynomial.rs
```diff
@@ -1737,395 +1776,270 @@ mod test_polynomials {
...
 
-    #[test]
-    fn fast_zerofier_test() {
-        let _1_17 = BFieldElement::from(1u64);
-        let _5_17 = BFieldElement::from(5u64);
-        let root_order: usize = 8;
+    #[proptest(cases = 50)]
+    fn zerofier_and_fast_zerofier_are_identical(
+        #[strategy(..4usize)] order_surplus: usize,
+        #[any(size_range(..1024).lift())] domain: Vec<BFieldElement>,
```
Sword-Smith, line 1790: This made me think what happens if there are repeated values in `domain`. This test still works, but does it mathematically give the right answer?

jan-ferdinand: Good question. In the cryptographic literature, I find definitions involving the minimality of a vanishing polynomial's[^0] degree, or equivalently, that the domain is a set. This conforms with my intuitive notion of a zerofier. More general texts are more relaxed in their definition, allowing arbitrary degree and arbitrary leading coefficients. Hence, we might want to either deny domains that are not sets, or discard duplicate elements.

[^0]: a different word for zerofier

jan-ferdinand: Due to the recursive nature of `fast_zerofier`, this might either mean a performance hit or a slight restructuring, with a uniqueness guaranteed established in some outer function and the recursion happening in an inner function.

PR #160, opened 2023-11-15
feat: return iterators in StorageVec trait

### LINKED ISSUE #159: StorageVec API is non-symmetric with regards to input and return types.  possible panic on 32 bit systems.
At present, the trait looks like:

```
pub trait StorageVec<T> {
    fn len(&self) -> Index;
    fn get(&self, index: Index) -> T;
    fn get_many(&self, indices: &[Index]) -> Vec<T>;
    fn get_all(&self) -> Vec<T>;
    fn set(&mut self, index: Index, value: T);
    fn set_many(&mut self, key_vals: impl IntoIterator<Item = (Index, T)>);
    fn set_all(&mut self, vals: &[T]);
    fn pop(&mut self) -> Option<T>;
    fn push(&mut self, value: T);
}
```

We can notice that `set` and `set_many` accept an index of type `Index` while `get_many` and `get_all` return a `Vec<T>`

`Index` is defined to be a `u64`.   `Vec` native indexes are of type `usize` which would be 2^32 on a 32 bit system and 2^64 on 64 bit system.   (to state the obvious)

In order to generate the return `Vec`, implementors of this trait cast the internal indexes of type `Index` to `usize`.   This cast can panic on a 32bit system if the index size is > usize::Max ie 2^32.   [example cast](https://github.com/Neptune-Crypto/twenty-first/blob/e0cf7e640162a730f56c409ce2b83cb111a339d8/twenty-first/src/util_types/storage_vec.rs#L175C13-L175C13).  So for example `get_all` would panic at the 2^32+1 element.  The situation is not quite as bad for `get_many`.  On a 32-bit system it would be limited to returning 2^32 elements, but would not panic (because the input list of indices is also an array/vec).

`set_all` also has a problem because it accepts `&[T]`.  On a 32 bit system this would be limited to 2^32 elements, which of course is much smaller than the 2^64 total addressable elements in StorageVec, eg as returned by `len()`.   For a collection with more than 2^32 values, `set_all()` would panic because implementors assert that input length is equal to target.len().

Some possible actions/remedies are:

1. no action.   leave as-is.    perhaps this could never be a problem in practice trying to get/set more than 2^32 values at once?
2. add doc-comments to warn caller, and leave responsibility with caller.
3. modify get_many, get_all, set_all to use (Index, T) instead of T.
4. leave `get_many`, `get_all`, `set_all`, but doc-comment them,  and add variants of each that uses `Vec<(Index, T)>` instead of `Vec<T>`.
5. define Index to be a usize rather than u64.

Of these, I prefer option 4.    It is backwards compatible with existing code/tests, but also provides a "more correct" path forward.  It leaves the existing methods for convenience, and places responsibility for length checking with the caller.

Option 1 seems not a good idea, especially as this is a crate published to crates.io.

Option 3 would require the most up-front work, and also there are many legitimate uses for calling the existing methods, where it is known that the length is less than 2^32.  And these methods are more convenient than passing around a tuple (Index, T).

Option 5 is simple.  It probably is not workable due to how this library is being used in practice, where more than 2^32 elements are expected in eg MMR.

dan-da:
This is a draft PR intended to show direction.   It is an attempt to address the issues raised in #159 and also enhance the API to support streaming very large datasets by returning iterators instead of allocating Vec.

Before merging, I intend to try out the API in neptune-core to make sure it is ergonomic.  It may yet need some tweaks.

Also, db locking is made granular again in get_many (and related).  I believe that's the only way it can work when returning an iterator instead of a Vec.  That code changed quite a bit, so I would like to benchmark old vs new.

I realize this is a bit of a departure from the way the StorageVec trait code was originally written.  I think/hope it should be an improvement, but will understand if the PR is not accepted.

The commit msg follows:

-----

close #159

This accomplishes three main objectives

 1. We return (Index, T) instead of T, making the get_* calls symmetric with the set_* calls.
 2. Values are returned as iterators instead of Vec<T> which means that it is possible for callers to iterate over very large sets without allocating a huge Vec.
 3. Maintains compatibility with existing code and tests. (except for very minor test changes for set_all)

storage_vec.rs:
 * get_many() is now just a convenience wrapper around get_many_iter()
 * add get_many_iter() to StorageVec trait
 * get_all() is now just a convenience wrapper around get_all_iter()
 * add set_first() call as convenience around set_many()
 * set_all() is a convenience around set_first() for when lengths must match
 * get_many_iter is now the work-horse, and is simpler than previous get_many()
 * add test get_many_iter_indexes

storage_schema.rs:
 * add get_many_iter() and re-work get_many logic to return an iterator
 * minor changse to test cases


Sword-Smith:
It's my experience with databases that you want as few calls to the database as possible. You want to batch-fetch everything you need outside of any loops, and the process the fetched data that lives in RAM inside of a loop and then write back any updates with a batch-write, thus keeping the number of database interactions as low as possible. From what I can see, the `get_many_iter` does the opposite: reads from the database one time for each element that it fetches.

I think this philosophy extends to locks: you don't want to hold and release a lock many times within a single RPC call, you want to grab the necessary locks, perform the update, and release them again. This not only removes locking/unlocking overhead but should also make deadlocks easier to resolve.

dan-da:
I don't disagree, but I prefer to test than theorize when possible.   :-)

To that end, I made a little test case that just pushes 10000 u64 into a dbtvec and then calls get_all() on that 1000 times.  It then iterates through the results, incrementing a counter to simulate an app using the data.   A second version of the test omits this result processing loop.

I ran these tests against release builds of the master branch and this PR branch.   I ran each test 3 times and picked the middle value.  The results are:

```
master branch:
  test_dbtcvecs_get_many_loop:  8.3 secs
  test_dbtcvecs_get_many_loop_no_iter_results: 8.35 secs. 

pr (iterator) branch:
  test_dbtcvecs_get_many_loop:  7.72 secs
  test_dbtcvecs_get_many_loop_no_iter_results: 0.01 secs. 
``` 
So not much of a difference between the two impls.  I believe both are reading everything from cache, and anyway this is an in-memory DB.  But at least we can see the iterator version is not slower for this simple test.   And further for the 2nd test that does not iterate the results, it is blazing fast due to the lazy nature of the iterator.   No data retrieval work is performed until the caller actually iterates.

I'm aware this test is simplistic and isn't really testing the case where data is present in DB but not in cache.  I'm attaching a [test case](https://github.com/Neptune-Crypto/twenty-first/files/13374242/test.txt).  If you have a suggestion how to make it more "real", I'd be happy to run them again.

Sword-Smith:
I modified your test case a bit, to also write the result back, and to persist. And to store the results on disk, and not on memory.

My results for this new test are:
This branch: 11.20s
`master`: 11.14

So it's a tie.

I run the test with this command:
`rm -rf twenty-first/test-database; cargo t time_db_interaction`

```rust
#[test]
    fn time_db_interaction() {
        let opt = rusty_leveldb::Options::default();
        // let opt = rusty_leveldb::default();
        let db = DB::open("test-database", opt.clone()).unwrap();

        // initialize storage
        let mut rusty_storage = SimpleRustyStorage::new(db);
        rusty_storage.restore_or_new();
        let mut vector = rusty_storage.schema.new_vec::<u64, u64>("test-vector");

        // Generate initial index/value pairs.
        const TEST_LIST_LENGTH: usize = 100000;
        for v in 0..TEST_LIST_LENGTH {
            vector.push(v as Index);
        }

        // Persist
        rusty_storage.persist();

        for _ in 0..10 {
            let mut v = vector.get_all();

            for v in v.iter_mut() {
                *v += 1;
            }

            // vector.set_all(v);
            vector.set_all(&v);

            rusty_storage.persist();
        }
    }
```

dan-da:
thx for the new test.   I tried it on both branches compiled with --release and I get between 3.2 and 3.5 seconds for each.   So yeah, no measurable difference.  I believe you got slower times because you are using a debug build.   I also commented out the `for _ in 0..10` loop temporarily so that only the initialization happens, and that took only 0.15 seconds, so that verifies most of the time is spent in the get_all/set_all loop.

dan-da:
I went down a little bit of a rabbit hole adding iter_mut() and many_iter_mut() support.  Whether we end up using it or not, it was, I think, a useful exercise for me learning more about mutating iterators and lending iterators which support windows/ranges and the kind of indirection we need for a trait.

I'm basically satisfied with this PR now, and ready to leave it alone.   ;-)

dan-da:
merged #167 instead.

PR #161, opened 2023-11-15
Remove itertools dep from bfieldcodec deriver

Sword-Smith:
- Removes itertools as dependency when deriving `BFieldCodec`
- Fixes a bug in calculating static encoding sizes when enum variants all have same data size

### DIFF
--- a/twenty-first/src/shared_math/bfield_codec.rs
+++ b/twenty-first/src/shared_math/bfield_codec.rs
@@ -1526,7 +1526,6 @@ mod tests {
 
         #[cfg(test)]
         mod simple_derivations {
-            use itertools::Itertools;
             use rand::random;
 
             use crate::shared_math::{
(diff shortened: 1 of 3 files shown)


twenty-first/src/shared_math/bfield_codec.rs
```diff
@@ -1524,9 +1525,56 @@ mod tests {
             }
         }
 
+        #[test]
+        fn encode_enum_with_uniform_data_size() {
```
jan-ferdinand, line 1529: Would it make sense to turn the two new tests into `#[proptest]`s?

jan-ferdinand (approved):

twenty-first/src/shared_math/bfield_codec.rs
Sword-Smith: Kind of. But I think we only need to run one test case here. More seems superfluous to me.

Sword-Smith:
Marking this as "draft" since @jan-ferdinand would like to merge all bfield-codec related code for the upcoming twenty-first release. He is working on bigger refactorings, so it makes sense he handles this.

PR #162, opened 2023-11-16
ignore duplicate elements in `zerofier` and `fast_zerofier`

jan-ferdinand:
There is at least one change in here that might be controversial:
I'm suggesting to change the interface of `fast_zerofier` to match that of `zerofier` by separating the recursed method and making it private.

### benchmarks

Comparing to current `master`.

#### `zerofier`

zerofier 0     
 time:   [12.896 ns 12.900 ns 12.906 ns]
                change: [+13.036% +13.432% +13.809%] (p = 0.00 < 0.05)
                Performance has regressed.

zerofier 10    
 time:   [717.83 ns 718.15 ns 718.52 ns]
                change: [+130.97% +131.13% +131.26%] (p = 0.00 < 0.05)
                Performance has regressed.

zerofier 100
    time:   [23.143 µs 23.158 µs 23.180 µs]
                change: [-17.408% -17.355% -17.302%] (p = 0.00 < 0.05)
                Performance has improved.

zerofier 1000  
 time:   [1.8308 ms 1.8312 ms 1.8317 ms]
                change: [-33.885% -33.844% -33.790%] (p = 0.00 < 0.05)
                Performance has improved.

zerofier 10000 
 time:   [175.44 ms 175.54 ms 175.64 ms]
                change: [-35.983% -35.940% -35.892%] (p = 0.00 < 0.05)
                Performance has improved.

#### `fast_zerofier`

fast_zerofier 0   
    time:   [32.621 ns 32.639 ns 32.661 ns]
                      change: [+131.87% +132.33% +132.76%] (p = 0.00 < 0.05)
                      Performance has regressed.

fast_zerofier 10      
time:   [1.9083 µs 1.9090 µs 1.9098 µs]
                      change: [+29.717% +29.824% +29.938%] (p = 0.00 < 0.05)
                      Performance has regressed.

fast_zerofier 100   
  time:   [34.669 µs 34.682 µs 34.697 µs]
                      change: [+10.837% +10.909% +10.982%] (p = 0.00 < 0.05)
                      Performance has regressed.

fast_zerofier 1000 
   time:   [590.28 µs 590.71 µs 591.15 µs]
                      change: [+6.5782% +6.6497% +6.7257%] (p = 0.00 < 0.05)
                      Performance has regressed.

fast_zerofier 10000 
  time:   [11.295 ms 11.299 ms 11.303 ms]
                      change: [+2.8685% +2.9211% +2.9727%] (p = 0.00 < 0.05)
                      Performance has regressed.


jan-ferdinand:
Given the benchmarks, we might also consider merging the two methods and introduce a cutoff point. `zerofier` is faster for small inputs, `fast_zerofier` is faster for big inputs.

Sword-Smith (changes requested):
The code looks good to me. And I like that you're using the `fast_zerofier_inner`  function as a helper function elsewhere and don't repeat the uniqueness check when you don't have to.

I guess the central question to me is whether we want to pay this performance penalty in exchange for changing the uniqueness responsibility from the caller to the callee. But maybe this performance hit could be avoided if you introduce the cutoff point mentioned above? 

I think we should try with the cutoff point and see if we can avoid the performance hit entirely. If there still is a performance hit for some small data sets, then so be it.

Sword-Smith (approved):
Let's not waste more time on this, since we're not really using it downstream anyway. If someone wants to squeeze 5 % more performance out of this, they can probably do that by adding the cutoff check to the recursive zerofier as well.

Issue #163, opened 2023-11-22
switch out rusty-leveldb for C++ impl.


dan-da:
This is a tracking issue for the task(s) of switching to C++ impl of levelb.

Some motivating factors for this are:

* The C++ impl supports concurrency so we should be able to eliminate many/most locks around DB access.
* The pure rust impl requires `&mut self` for read calls such as the `get()` method.  This requires that we use mutex locks, which makes readers block for eachother, and writer also must block for readers.  
* The C++ impl is used a lot more widely, including in bitcoin-core, and so is much more mature and (arguably) better maintained.
* The C++ impl may be faster.  (untested).
* It is hoped/expected we will have better stability with fewer locks and a more mature DB.
* The C++ impl has built-in cache, so we may not need our own cache layer.  (TBD)

A challenge is that the C++ impl does not support fully in-memory databases.  So on-disk database directory is created for each DB, and must somehow be cleaned up for tests, including tests that panic: on purpose or not.

Rough Roadmap:

- [x] evaluate which rusty leveldb crate to use.  (chose rs-leveldb because it supports more key types)
- [x] switch dep from rusty-leveldb to rs-leveldb and make code/tests compile.
- [x]  implement auto-destruction of database directories for tests
- [x] get all tests passing without any test database directories persisting after test run.
- [x] verify that neptune-core still passes all tests also.
- [x] document deps to build with c++ impl.
- [x] milestone:  parity with rusty-leveldb based code.   API mostly unchanged.
- [x] remove unnecessary locks in codebase
- [x] remove unnecessary locks in tests
- [x] time test(s) and/or create benchmark tests.  compare with rusty-leveldb codebase.
- [x] evaluate if we can remove our cache(s)
- [x] evaluate further simplifications
- [x] consider/evaluate moving all types that utilize the DB into a separate crate.

Sword-Smith:
I'm excited about this because it can reduce the number of locks, and because we switch to a well-tested implementation! If you have the chance, I would love to see one or two benchmarks comparing performance.

dan-da:
[...]

PR #164, opened 2023-11-22
feat: use rs-leveldb instead of rusty-leveldb

dan-da:
Summary: This is a necessary first step that just performs a straight swap of leveldb crates but keeps our existing locks in place and changes as little code as practical.   We could review/merge this now, or wait until I've removed some locks.  (which will be a bigger set of diffs)

--------

Addresses #163

rs-leveldb is a wrapper for C++ leveldb.  It is a fork of the standard leveldb crate and seems nicer because it supports using strings and standard integer types as keys directly.

With these changes the code is equally as functional and all tests pass. Application locks remain in place, but are redundant to the locking that occurs within C++ leveldb.  Our present Mutex locks harm concurrency, so it is planned that a future commit will remove these.

Completes tasks:

- [x] switch dep from rusty-leveldb to rs-leveldb and make code/tests compile. 
- [x] implement auto-destruction of database directories for tests 
- [x] get all tests passing without any test database directories persisting after test run. 
- [x] document deps to build with c++ impl.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,3 +44,11 @@ While twenty-first's version is `0.x.y`, releasing a new version:
 [crates]: https://crates.io/crates/twenty-first/versions
 
 If you do not have the privilege to create git tags or run `cargo publish`, submit a PR and the merger will take care of these.
+
+## Building
+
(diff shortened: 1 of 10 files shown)


Sword-Smith:
@jan-ferdinand is preparing the release of version 0.36 which [Triton-VM](https://github.com/TritonVM/triton-vm) will depend on. So let's prioritize getting the DB and DB wrappers in as good a state as possible, so the entire dependency chain can benefit from this.

dan-da:
merged #167 instead.

PR #165, opened 2023-11-23
overhaul `BFieldCodec` and associated derive macro

jan-ferdinand:
- [x] flip encoding order for tuples, structs, and enums' associated data: last field is encoded first
- [x] remove `BFieldCodec`-derived code's dependency on `itertools`
- [x] remove `BFieldCodec`-derived code's dependency on `anyhow`, derive custom error types instead
- [x] make the error type of `BFieldCodec` an associated type
- [x] use `proptest` framework for `BFieldCodec` property tests
- [x] always run all property tests: remove conditional execution based on environment variable
- [x] greatly reduce necessary boilerplate to property-test any type for `BFieldCodec` properties
- [x] add `BFieldCodec` deriving for unit structs[^0]

supersedes #161

[^0]: Note that deriving `BFieldCodec` for `struct MyStruct {}` and `struct MyStruct();` was already possible. This addition just extends that to `struct MyStruct;`.


aszepieniec:
> I'm assuming it only matters that the respective published versions of the various libraries are updated accordingly, correct?

It is imperative that the most recent published versions match. It would be nice to have the matching branch of `tasm-lib` ready ASAP after this one is deemed ready, so that we can anticipate the updates when writing snippets.

Sword-Smith (approved):

PR #166, opened 2023-11-28
Leveldb switch/refactor 2nd pass

dan-da:
I have a few more changes/tweaks to make tomorrow, but this draft PR gives a good idea of where its going.

No need for a review yet.

I will update this description with full summary of changes and some benchmark results later/tmw.


dan-da:
I started making tests for atomic batch writes and reads for RustyLevelDBVec and discovered that RustyLevelDbVec, DbtVec, etc could not be shared between threads in the same way as Arc<RwLock<..>> can be.  Even though the former are now just NewType wrappers for the latter.  One problem was methods unnecessarily taking `&mut self..  Some big-ish refactors later, and the concurrency tests work.  So it took longer than planned to get to this point, but its better to catch such issues now before publishing the crate.

I still need to make concurrency tests for DbtVec, DbtSingleton and multiple StorageSchema tables tomorrow, but I expect (fingers crossed) those should go smoothly.

dan-da:
merged #167 instead.

PR #167, opened 2023-12-11
Leveldb switch and storage API refactor.

dan-da:
This PR is ready to begin review, with the following caveats:

1. I plan to add a more detailed description of changes and points for review/discussion tmw.
2. Perhaps some commits should be squashed, commit msg improved etc.  There are lots of (wip: work-in-progress) commits with terse messages, and also some commits that do not pass tests, etc, as I explored many avenues.
3. I have not yet investigated or impl'd any support for `cargo-dist`, to facilitate building C++ deps.  That may block this from being merged. 

For these reasons, I am marking this as a draft PR, for now.

At a very high level, this PR does the following:

* replaces `rusty-leveldb` (rust) with `rs-leveldb`, a C++ impl.  Obsoletes #164.
* refactors, simplifies, extends storage API substantially and moves it to `crate::storage`. Obsoletes #166.
* adds `crate::sync::{AtomicRw, AtomiMutex}` wrappers that simplify usage of `Arc<RwLock>` and `Arc<Mutex>`.
* changes storage lock types from `Mutex` to `RwLock`
* adds concurrency tests to verify atomicity of types in `crates::storage`
* adds bench tests for levelDB direct access as well as `DbtVec`.
* implements a flexible de/serialization strategy/API for `storage` types.
* adds detailed doc-comments and descriptions for all types in `storage` and `sync` as well as design notes for `storage`.

Patches are necessary to make `neptune-core` build against this branch.  I have verified that a patched `neptune-core` passes all tests and also correctly loads existing wallet and blockchain databases and is able to mine blocks.

The `neptune-core` diffs are [here](https://github.com/Neptune-Crypto/neptune-core/compare/master...dan-da:neptune-core:leveldb_switch). 

@Sword-Smith This is a big PR and I imagine you will have many q's/comments.  It may be best to meet and I can give you an overview of changes and rationale before starting review.


dan-da:
One of the CI test runs had two failures of the non-atomic tests that should panic:

```
failures:
  storage::storage_schema::dbtvec::tests::non_atomic_set_and_get    
  storage::storage_schema::tests::concurrency::multi_table_storage_vec::non_atomic_multi_table_setmany_and_getmany
```

These tests are not deterministic.   The test passes when a panic occurs due to an inconsistency in the reader thread. 

If this continues happening, I can add (a) more iterations, (b) write order randomization, (c) timing randomization (sleeps) between threads to make the inconsistencies appear sooner.      Or worst case, we can disable those tests now that they've served their initial purpose.

dan-da:
I marked this PR ready for review.

The biggest concern was about building/installing the C++ leveldb.   It turns out that this is handled by the [leveldb-sys crate](https://crates.io/crates/leveldb-sys/2.0.9) which `rs-leveldb` depends on.

[Here](https://matt-harrison.com/posts/rust-sys-crate/) is a general description of *-sys crates.  [and another](https://doc.rust-lang.org/cargo/reference/build-scripts.html#-sys-packages)   The tl;dr is that they take care of ensuring an external lib exists or building it if not, in a cross platform way.

So it "just works" and we don't need to worry about it.

Sword-Smith:
> I marked this PR ready for review.


Very nice. Thanks! We are still struggling with the porting of [`tasm-lib`](https://github.com/TritonVM/tasm-lib) to Triton VM v0.36, so we haven't updated the entire dependency chain yet. That makes me think that this PR should be part of the next `twenty-first` major version, and not just the subsequent one as we earlier had envisioned.

.vscode/settings.json
Sword-Smith, line 1: Should we maybe add this file to `.gitignore`? I'm not sure it belongs here.

twenty-first/src/storage/storage_vec/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    // todo: delete fn
+    pub(super) fn get_test_db(destroy_db_on_drop: bool) -> Arc<DB> {
+        Arc::new(DB::open_new_test_database(destroy_db_on_drop, None, None, None).unwrap())
+    }
+
+    // todo: delete fn
+    fn open_test_db(path: &std::path::Path, destroy_db_on_drop: bool) -> Arc<DB> {
+        Arc::new(DB::open_test_database(path, destroy_db_on_drop, None, None, None).unwrap())
+    }
```
Sword-Smith, line 44: Why do you want to delete these helper functions?

twenty-first/src/storage/storage_schema/mod.rs
```diff
@@ -0,0 +1,1482 @@
...
+        use super::*;
+        use crate::sync::AtomicRw;
+        use std::thread;
+
+        type TestVec = DbtVec<u64>;
+        type TestSingleton = DbtSingleton<u64>;
+
+        // pub fn prepare_concurrency_test_singleton(singleton: &impl StorageSingleton<u64>) {
+        //     singleton.set(42);
+        // }
```
Sword-Smith, line 989: Delete?

Sword-Smith (changes requested):
This looks very promising. I like have you can use closures to encapsulte the holding of the lock.

I have some questions though:
1. This changes the lock type from `std::sync::Mutex` to `std::sync::RwLock`? Why is that better?
2. leveldb can verify some checksums. I don't think you do that here, as the default value is `false`. Should we do that?
3. Do we have any benchmarks comparing our previous crate `rusty_leveldb` with the new crate `leveldb`?
4. The `simple_rusty_*` names came from the `rust_leveldb` crate name, I think. Maybe we want to change that name?

twenty-first/src/storage/storage_schema/simple_rusty_storage.rs
```diff
@@ -0,0 +1,63 @@
...
+/// possible to use this struct and add to the schema.
+pub struct SimpleRustyStorage {
+    /// dynamic DB Schema.  (new tables may be added)
+    pub schema: DbtSchema<SimpleRustyReader>,
+}
+
+impl StorageWriter for SimpleRustyStorage {
+    #[inline]
+    fn persist(&mut self) {
+        let write_batch = WriteBatch::new();
```
Sword-Smith, line 20: It's very mysterious to me that `write_batch` does not have to be mutable. I looked through the call graph and saw that it's because `unsafe` is used internally, to call the c++ function, I think. Should we be worried about that use of `unsafe`?

Sword-Smith:
I added a few commits here that you can merge into this PR:
https://github.com/Neptune-Crypto/twenty-first/tree/pr167-thor

.vscode/settings.json
dan-da: yes, I didn't intend to commit it.  I will add to .gitignore.

twenty-first/src/storage/storage_vec/mod.rs
dan-da: I think that was because before they were doing something a bit more complicated, but now it's mostly done in DB::open_test_database().   They are probably still a fine shorthand though, so I may just drop the comment.

dan-da:
> This looks very promising. I like have you can use closures to encapsulte the holding of the lock.

yeah, I like that the closures make it super clear when a lock is acquired and dropped.

A subtle thing is that if one always uses the closures, one can switch back and forth between AtomicRw and AtomicMutex and it will "just work" in either direction.  Whereas using the lock_guard* methods, one can always go from AtomicRw to Mutex, but not necessarily the reverse.  (Because the MutexGuard returned by AtomicMutex::lock_guard() is actually mutable ,so callers might (mistakenly) use it that way).  

> I have some questions though:
> 
> 1. This changes the lock type from `std::sync::Mutex` to `std::sync::RwLock`? Why is that better?

This is an area of tradeoffs.   Ultimately to know which is better, we will need to benchmark for particular use cases we care about.

We can generalize that Mutexes are a bit faster than RwLock.   However:
* Mutex read operations block eachother
* RwLock read operations run concurrently.  They are only blocked by writer.

So, for any read-only methods, it is more concurrent to implement them as RwLock read ops.

In general my approach is to make write lock acquisitions as short and simple as possible.   So if I have a method that does a bunch of read-only processing at the top and then writes at the end, I will do the top stuff with a read lock, and then acquire the write lock only for the actual write.

If we have some use-case where there are multiple threads/tasks doing mainly write ops, and we make the locked sections as short as possible then mutex may be better/faster.  But I think for most use cases RwLock will behave more concurrently.

So yeah, I am defaulting to RwLock and figure we can bench and switch to Mutex for specific cases if/when performance issues reveal themselves.

> 2. leveldb can verify some checksums. I don't think you do that here, as the default value is `false`. Should we do that?

I think that makes sense yes.  I will change the default.   In benches it added some overhead, but not that much.    It seems a good safety feature to have on during development especially.

> 3. Do we have any benchmarks comparing our previous crate `rusty_leveldb` with the new crate `leveldb`?

yes, I backported the benches to master and ran.  I thought I had put the results into the previous PR or issue, but I don't see them now.   I will look around.   Anyway, my recollection is that the numbers were pretty similar.

> 4. The `simple_rusty_*` names came from the `rust_leveldb` crate name, I think. Maybe we want to change that name?

Yeah.... I did find some of those curious.  I've been keeping the naming pretty much the same to avoid any more breakage than necessary.   Do you have name suggestions?

> 5. The [`rusty_leveldb`](https://crates.io/crates/rusty-leveldb) crate that we used previously is much more popular than this new crate, [`rs-leveldb`](https://crates.io/crates/rs-leveldb). Should we be concerned about that?

short(ish) answer:
1. `leveldb` crate is C++, and has 900,000+ uses vs 50,000+ uses for `rusty-leveldb`
2. `rs-leveldb` is a fork of `leveldb` that has support for accepting most primitive types as keys.
3. `rs-leveldb` only has ~2000+ uses.
4. Getting started, I found it much cleaner/easier to use the rs-leveldb interface.  Its the same C++ code underneath, and mostly the same Rust code as `leveldb`.
5. If we have problems, it shouldn't be too big of a deal to switch to `leveldb` or even back to `rusty-leveldb`.   (now that my rust and leveldb understanding have leveled-up.)

twenty-first/src/storage/storage_schema/simple_rusty_storage.rs
```diff
@@ -0,0 +1,63 @@
...
+        self.schema.tables.lock(|tables| {
+            for table in tables.iter() {
+                let operations = table.pull_queue();
```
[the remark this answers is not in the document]
dan-da: I will add a doc-comment.

twenty-first/src/storage/storage_schema/mod.rs
dan-da: thx.

twenty-first/src/storage/storage_schema/simple_rusty_storage.rs
dan-da: unsafe is used internally in both `leveldb` and `rs-leveldb`.   It's not unusual when dealing with C pointers and such in rust.

However, it is interesting to note that in `leveldb` WriteBatch::put() takes `&mut self`, but in `rs-leveldb` it takes `&self`.    I would need to dig into the projects commit history to (maybe) understand why.

The biggest thing that concerns me about the `rs-leveldb` crate so far is that when I was making the bench tests, I was getting some segfaults, I believe when writing.    I figured out this only occurred when passing a non default Cache object in the config `Options`, so I stopped doing that, and made a mental note to look into it later.  Haven't seen a segfault or other problem since.  Now that you point out this discrepancy, well it could possibly be related somehow.

My feeling is that we could merge this PR with `rs-leveldb`, and change it a bit later if decide we need to.  We aren't in production yet.   And rs-leveldb has otherwise been solid in my test usage.   I'm not sure how long it would take to switch to `leveldb`, but perhaps a couple days.

Sword-Smith: Alright. Let's stickwith the default cache in options then

Sword-Smith:
> In general my approach is to make lock acquisitions as short and simple as possible. _especially write-lock blocks_. So if I have a method that does a bunch of read-only processing at the top and then writes at the end, I will do the top stuff with a read lock, and then acquire the write lock only for the actual write(s).

If the thing that is being written back depends on the integrity of the read data, don't we need to hold a lock a write lock over the processing of the data, such that we ensure that the read data corresponds to what's in the database now? I think that's true for the general case, but for many of these data structures only the `main_loop` thread actually writes, so it might not be a problem. What do you think?

Sword-Smith:
I was thinking `storage/database_array.rs`, `storage/database_vector.rs` could be deleted. 

Also, maybe `storage/storage_vec/ordinary_vec.rs` and `storage/storage_vecordinary_vec_private.rs` could be put behind a `#[cfg(test)]` annotation since they are only used for tests. This 2nd suggestion led to failing doc-tests though.

If you prefer, this cleanup can also be handled at a later stage.

dan-da:
I added some simple benches for AtomicRw and AtomicMutex lock acquisitions.

```
Note that:
 1. `lock` and `lock_guard` denote read-lock acquisitions
 2. `lock_mut` and `lock_guard_mut` denote write-lock acquisitions
 3. For mutex, only write-lock acquisitions are possible.

sync_atomic              fastest       │ slowest       │ median        │ mean          │ samples │ iters
├─ lock                                │               │               │               │         │
│  ╰─ rw                               │               │               │               │         │
│     ╰─ lock_guard      169.1 µs      │ 210.6 µs      │ 169.1 µs      │ 175.3 µs      │ 100     │ 100
╰─ lock_mut                            │               │               │               │         │
   ├─ mutex                            │               │               │               │         │
   │  ╰─ lock_guard_mut  131.8 µs      │ 217.9 µs      │ 131.8 µs      │ 136.5 µs      │ 100     │ 100
   ╰─ rw                               │               │               │               │         │
      ╰─ lock_guard_mut  131.8 µs      │ 153.8 µs      │ 131.8 µs      │ 132.7 µs      │ 100     │ 100

Analysis:
 1. RwLock and Mutex write-lock acquisitions are basically the same.
 2. RwLock read-lock acquisitions are about 22% slower than write-lock acquisitions
    which seems acceptable for most uses.
```

.vscode/settings.json
dan-da: ok, I git rm'd .vscode and added .vscode to gitignore.   This way devs can do their own thing.

twenty-first/src/storage/storage_vec/mod.rs
dan-da: comments removed.

twenty-first/src/storage/storage_schema/simple_rusty_storage.rs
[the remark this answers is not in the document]
dan-da: added doc-comment to DbtVec::pull_queue().    The cache is impl detail of DbtVec, so does not apply to eg DbtSingleton or possible future types.

twenty-first/src/storage/storage_schema/mod.rs
dan-da: deleted.

dan-da:
> If this continues happening, I can add (a) more iterations, (b) write order randomization, (c) timing randomization (sleeps)

I added a commit with iterations bumped (1,000 -> 10,000) and also added a 1 ms sleep in the reader thread.  The tests pass reliably on my machine.... hopefully they will on CI also.

dan-da:
> If the thing that is being written back depends on the integrity of the read data, don't we need to hold a lock a write lock over the processing of the data, such that we ensure that the read data corresponds to what's in the database once we do the writing? I think that's true for the general case, but for many of these data structures only the main_loop thread actually writes, so it might not be a problem. What do you think?

I think you make a good point. 

It's good to keep in mind there is a distinction between a `critical section` and `atomicity` of data.   I find [this discussion](https://softwareengineering.stackexchange.com/questions/418901/do-critical-section-and-atomicity-imply-each-other) helpful in thinking about it.

In a scenario with multiple writers, my approach above can lead to data being written that is based on an out-dated view/snapshot.  When another writer thread sneaks in between "my" read and write locks.

If there is just a single writing thread, then that approach should be more concurrent, with single writer, many readers.

I will review further in neptune-core exactly what threads are being started and what roles they play, and doc-comment what I find, to make it clearer if the lock usage in my refactors is atomic or not.

So yeah, that is a good thing to review in Neptune and level-up my understanding there.  I don't think it affects this PR though.

dan-da:
> I added a commit with iterations bumped (1,000 -> 10,000) and also added a 1 ms sleep in the reader thread. The tests pass reliably on my machine.... hopefully they will on CI also.

Dang!   CI test is still failing.    So for now I've commented out the 2 non-deterministic sometimes-failing tests.   Hopefully by adding randomness I can make them more reliable in the future, but I don't have any more time to work on this PR today.

dan-da:
Ok, I think this PR is ready for final review/merge.  Let me know if any concerns.     

I will see about getting the commented tests working on CI reliably later on this week.

dan-da:
Touching on a couple things I missed in my commits:

> leveldb can verify some checksums. I don't think you do that here, as the default value is false. Should we do that?

Upon review, I think we should do that in Neptune, but here in storage we just (a) wrap levelDB API and (b) use an instantiated DB that is given to us.   So I don't think a change is needed in twenty-first, but rather the app(s) should set that option.

> The simple_rusty_* names came from the rust_leveldb crate name, I think. Maybe we want to change that name?

I'm out of time/bandwidth for today.   I can do some name changing tmw if that's ok.  Or otherwise I'm fine with keeping them as-is for the upcoming release.

dan-da:
> I was thinking `storage/database_array.rs`, `storage/database_vector.rs` could be deleted.
> 
> Also, maybe `storage/storage_vec/ordinary_vec.rs` and `storage/storage_vecordinary_vec_private.rs` could be put behind a `#[cfg(test)]` annotation since they are only used for tests. This 2nd suggestion led to failing doc-tests though.
> 
> If you prefer, this cleanup can also be handled at a later stage.

yeah, let's do it later.  Small thing, but I'm a bit overloaded this week.

Sword-Smith (approved):
There are some types that could be removed and some modules that can be renamed, as mentioned about, but I'm happy to merge this now.

dan-da:
> If the thing that is being written back depends on the integrity of the read data, don't we need to hold a lock a write lock over the processing of the data, such that we ensure that the read data corresponds to what's in the database once we do the writing? I think that's true for the general case, but for many of these data structures only the main_loop thread actually writes, so it might not be a problem. What do you think?

Following up on this.

In the general case, it works to first acquire a mutex over unrelated data (eg `()`) and hold it across both the read and write locks over `mydata`.

With recent versions of stable rust, this can be accomplished with const Mutex.    eg:

```
use std::sync::{Arc, RwLock, Mutex};

pub fn read_and_write(shared_data: Arc<RwLock<bool>>) {

    static ATOMIC_MUTEX: Mutex<()> = Mutex::new(());
    
    {
        let atomic_update_section = ATOMIC_MUTEX.lock().unwrap();
        
        {
            let read_guard = shared_data.read().unwrap();
            // perform lenghthy read ops
            drop(read_guard)
        }
        
        {
            let write_guard = shared_data.write().unwrap();
            // perform write ops
            drop(write_guard);
        }
        
        drop(atomic_update_section);
    }
}
```
[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1456ce93bbe107df93b0b9d2ad26b815)

note: tokio Mutex also supports const usage with Mutex::const_new().

The benefit of this pattern is that other threads can concurrently read the data while `read_and_write` is performing its read operations.   Yet any write threads that call `read_and_write` will have to atomically perform the read and write because they must first acquire the `atomic_update_section` mutex.

This could be done with a regular non-const Mutex, but then that Mutex must be stored outside `read_and_write` and its not useful for anything else.

PR #169, opened 2023-12-16
(de)serialize `BFieldElement` in canonical representation

jan-ferdinand:
Previously, (de)serialization used the internal, i.e., Montgomery representation. Now, deserializing (for example) a “1” gives the `BFieldElement` 1.

The motivation for this change comes from the [debugger for Triton VM](https://github.com/TritonVM/triton-vm/pull/242). Concretely, when specifying digests for non-deterministic input, they currently have to be specified with the components in Montgomery representation. I believe this to be confusing, as the digest appearing on the VM's stack will seemingly differ from the specified digest.


Sword-Smith (approved):
Good tests. LGTM.

aszepieniec (approved):
LGTM

PR #171, opened 2023-12-26
improve `MerkleTree` implementation

### LINKED ISSUE #134: Populate Triton VM memory with a Merkle tree
With the release of 0.30.0 the Merkle tree nodes were made private. But some code in `tasm-lib` relied on those node digests being accessible, specifically

```rust
for (i, node) in tree.nodes.into_iter().enumerate().skip(1) {
            for j in 0..DIGEST_LENGTH {
                memory.insert(
                    list_address
                        + BFieldElement::one()
                        + BFieldElement::new((i * DIGEST_LENGTH + j) as u64),
                    node.values()[j],
                );
            }
        }
```

One option for rewriting this code is to have a `get_all_nodes` function. Another option is to implement `BFieldCodec` for the Merkle tree structure. The latter option is probably preferable, as that would ensure that the Merkle tree is put into memory in a canonical representation.

jan-ferdinand:
- [x] don't panic, return `Err(…)`
- [x] remove prefix `get_` from methods' names
- [x] rely on built-in methods more
- [x] allow (read-only) access to internal nodes (fix #134)
- [x] make all `Arbitrary` instances legal
- [x] use `proptest` framework
- [x] assess if further refactors are beneficial
  - [x] reduce length of argument lists
  - [x] introduce additional `struct`s, in particular, `MerkleTreeInclusionProof` and `PartialMerkleTree`

### DIFF
--- a/twenty-first/src/util_types/merkle_tree.rs
+++ b/twenty-first/src/util_types/merkle_tree.rs
@@ -11,7 +11,6 @@ use rayon::iter::{IndexedParallelIterator, IntoParallelIterator, ParallelIterato
 use thiserror::Error;
 
 use crate::shared_math::digest::Digest;
-use crate::shared_math::other::{is_power_of_two, log_2_floor};
 use crate::util_types::algebraic_hasher::AlgebraicHasher;
 use crate::util_types::merkle_tree_maker::MerkleTreeMaker;
 
@@ -351,14 +350,14 @@ where
 
     pub fn num_leafs(&self) -> usize {
         let node_count = self.nodes.len();
-        assert!(is_power_of_two(node_count));
+        assert!(node_count.is_power_of_two());
         node_count / 2
     }
 
     pub fn height(&self) -> usize {
-        let leaf_count = self.num_leafs() as u128;
-        assert!(is_power_of_two(leaf_count));
-        log_2_floor(leaf_count) as usize
+        let leaf_count = self.num_leafs();
+        assert!(leaf_count.is_power_of_two());
+        leaf_count.ilog2() as usize
     }
 
     /// All leaves of the Merkle tree.
@@ -388,7 +387,7 @@ impl<H: AlgebraicHasher> MerkleTreeMaker<H> for CpuParallel {
         let leaves_count = digests.len();
 
         assert!(
-            is_power_of_two(leaves_count),
+            leaves_count.is_power_of_two(),
             "Size of input for Merkle tree must be a power of 2"
         );
 
@@ -913,7 +912,7 @@ pub mod merkle_tree_test {
         let exponent = 6;
         let num_leaves = usize::pow(2, exponent);
         assert!(
-            is_power_of_two(num_leaves),
+            num_leaves.is_power_of_two(),
             "Size of input for Merkle tree must be a power of 2"
         );
 
@@ -964,7 +963,7 @@ pub mod merkle_tree_test {
         let exponent = 6;
         let num_leaves = usize::pow(2, exponent);
         assert!(
-            is_power_of_two(num_leaves),
+            num_leaves.is_power_of_two(),
             "Size of input for Merkle tree must be a power of 2"
         );
 
--- a/twenty-first/src/util_types/merkle_tree.rs
+++ b/twenty-first/src/util_types/merkle_tree.rs
@@ -583,6 +583,14 @@ pub mod merkle_tree_test {
     }
 
     #[proptest]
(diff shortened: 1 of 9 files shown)


twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -31,30 +34,36 @@ where
...
-/// These are exactly the arguments for the `verify_*` family of static methods.
 impl<H> MerkleTree<H>
 where
     H: AlgebraicHasher,
 {
-    /// Given a list of leaf indices, return the indices of exactly those nodes that are needed to
-    /// prove (or verify) that the indicated leaves are in the Merkle tree.
-    // This function is not defined as a method (taking self as argument) since it's
-    // needed by the verifier who does not have access to the Merkle tree.
+    const MAX_NUM_LEAVES: usize = usize::MAX / 2;
```
Sword-Smith, line 41: I guess in the Triton VM context, the max number of leaves is $2^{31}$. And it should probably always be a power of two which is not enforced here, no?

jan-ferdinand: Good points. :+1: I think setting it to $2^{31}$ makes sense. Depending on the STARK's security parameters, the actual number of nodes usable in Triton VM might be fewer; however, in my opinion, twenty-first should not concern itself with that level of detail.

Sword-Smith:
Should we allow Merkle trees to be created from an empty list of digests? This code panics when that is attempted.

Our MMR implementation needs to be able to handle empty lists, but I don't think our Merkle tree implementation needs to be able to handle that.

I'm OK with not allowing that, the `digests` argument to `from_digests` be empty. In that case, we should probably remove the `root_from_arbitrary_number_of_digests_empty_test` test? I'm a bit unsure whether to leave that test.

I would also like to see a test that explicitly handles inputs of length 0 and 1. See the branch `thv/merkle-tree` for such tests. Feel free to change them as you see fit.

jan-ferdinand:
> Should we allow Merkle trees to be created from an empty list of digests? This code panics when that is attempted.
> 
> Our MMR implementation needs to be able to handle empty lists, but I don't think our Merkle tree implementation needs to be able to handle that.

In my mind, a Merkle tree has $2^n$ leaves[^1]; anything else is a Merkle Mountain Range. I'd prefer that behavior, but am ready to yield if there are counter arguments.

> I'm OK with not allowing that, the `digests` argument to `from_digests` be empty. In that case, we should probably remove the `root_from_arbitrary_number_of_digests_empty_test` test? I'm a bit unsure whether to leave that test.

I think the function `root_from_arbitrary_number_of_digests` and all related tests fit better with our MMR stuff.

> I would also like to see a test that explicitly handles inputs of length 0 and 1. See the branch `thv/merkle-tree` for such tests. Feel free to change them as you see fit.

Good point. However, we need consensus on the expected behavior first, of course. :relieved:

[^1]: also implying a minimum of one leaf

Sword-Smith:
> In my mind, a Merkle tree has 2n leaves[1](#user-content-fn-1-a0a3108590e17e0d3c628a6e94b49a19); anything else is a Merkle Mountain Range. I'd prefer that behavior, but am ready to yield if there are counter arguments.

I'm happy with this definition. But I guess `from_digests` should return an error if called with an empty slice/list then, and not panic as it does now?

jan-ferdinand:
> I guess `from_digests` should return an error

I fully concur. f584a6115f77a97c0bac517ed5bd9f2dbc77dbce proposes the corresponding change.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -28,10 +27,16 @@ const MAX_NUM_NODES: usize = 1 << 32;
 const MAX_NUM_LEAVES: usize = MAX_NUM_NODES / 2;
 pub const MAX_TREE_HEIGHT: usize = MAX_NUM_LEAVES.ilog2() as usize;
 
-const ROOT_INDEX: usize = 1;
+/// The index of the root node in a [Merkle tree](MerkleTree).
+/// It is recommended to use [`root()`](MerkleTree::root) instead.
+pub const ROOT_INDEX: usize = 1;
```
Sword-Smith, line 32: Why make this public?

Sword-Smith (approved):
I would not make `ROOT_INDEX` public. And I would add corner case tests of `tree.height()` and `tree.num_leaves()` public functions, for a tree with only one leaf.

Sword-Smith (approved):
I would not make `ROOT_INDEX` public, and I would add tests of `num_leafs` and `height` methods for a Merkle tree with one leaf.

PR #172, opened 2024-01-02
Thv/merkle tree

Sword-Smith:
- Add two tests -- one of which are currently failing
- Reduce scope of a constant used for Merkle tree building
- Add a constant for the root index


jan-ferdinand:
Thanks for the suggestions. :ok_hand: I've incorporated what I think fits with the re-structure of #171. Closing for now, but happy to discuss further over there once that is ready for review.

PR #173, opened 2024-01-08
Lock notifications pr

dan-da:
This PR consists of two commits that:

1) Adds an optional lock name and event callback fn to `AtomicRw` and `AtomicMutex`, which enables applications to log/trace all lock events  (TryAcquire, Acquire, Release).

2) makes use of the event callback functionality within the `Storage` types, again in a way that is optional for the application.

**Breaking Changes:**

There is one breaking API change that I'm aware of.   I added a field to `DbtSchema` which previously could only be created by direct struct {..} instantiation.   So there is no way to add a field in a non-breaking way.   This pr also adds DbtSchema::new(). 

**Motivation:**

I wanted to be able to log/trace all lock events in neptune-core, to be able to track which threads and tokio tasks are using each lock and in what manner.   For this, each lock needs to have a unique name.  

**Design Decisions:**

*How to impl Release event*.   A lock is not released until the corresponding lock guard is dropped.  For the `lock_guard()` and `lock_guard_mut()` methods this is a problem because the caller directly holds the guards.  I found a solution by wrapping the guard with a wrapper eg `AtomicMutexGuard`, which impls `Deref` and `DerefMut`.  So now the caller uses the wrapper just as they would a guard from the std lib.

*Callback or not.*  It perhaps would have been simpler to just log/trace within twenty-first without using any callback function.  I decided to go with the callback approach because:
* I wanted to also log tokio task identifiers, and twenty-first has no dep on tokio.
* I didn't want to make any assumptions about application's log/trace facilities or output format
* this just seems a more flexible approach.

*modify storage API or not*.  It might be cleaner to pass the lock name to `DbtSchema::new_vec()` and `DbtSchema::new_singleton()` however that would be a breaking change.  Instead I used existing parameters and the `DbtSchema` name to auto generate a lock name.

**Reviewing:**

I suggest to start review by looking at the changes in crate::sync, followed by changes in crate::storage.   The latter wil make more sense that way.

**Example**

Example log output that this PR enables, from a callback fn in neptune-core:

```
2024-01-08T17:42:06.746452Z TRACE block_template_is_valid_test: neptune_core: TryAcquire lock `RustyArchivalMutatorSet-Schema-DbtVec - aocl` of type `RwLock` for `Read` by
        |-- thread 74, (`mine_loop::mine_loop_tests::block_template_is_valid_test`)
        |-- tokio task ?
        |-- lock_event=TryAcquire { info: LockInfo { name: Some("RustyArchivalMutatorSet-Schema-DbtVec - aocl"), lock_type: RwLock }, acquisition: Read }
2024-01-08T17:42:06.746613Z TRACE block_template_is_valid_test: neptune_core: Acquire lock `RustyArchivalMutatorSet-Schema-DbtVec - aocl` of type `RwLock` for `Read` by
        |-- thread 74, (`mine_loop::mine_loop_tests::block_template_is_valid_test`)
        |-- tokio task ?
        |-- lock_event=Acquire { info: LockInfo { name: Some("RustyArchivalMutatorSet-Schema-DbtVec - aocl"), lock_type: RwLock }, acquisition: Read }
2024-01-08T17:42:06.746686Z TRACE block_template_is_valid_test: neptune_core: Release lock `RustyArchivalMutatorSet-Schema-DbtVec - aocl` of type `RwLock` for `Read` by
        |-- thread 74, (`mine_loop::mine_loop_tests::block_template_is_valid_test`)
        |-- tokio task ?
        |-- lock_event=Release { info: LockInfo { name: Some("RustyArchivalMutatorSet-Schema-DbtVec - aocl"), lock_type: RwLock }, acquisition: Read }
```

(note: tokio task id is not displaying in neptune-core tests for some reason and displays as "?".  That's unrelated to this PR.)

### DIFF
--- a/twenty-first/src/storage/storage_schema/dbtsingleton.rs
+++ b/twenty-first/src/storage/storage_schema/dbtsingleton.rs
@@ -3,7 +3,7 @@ use std::{fmt::Debug, sync::Arc};
 use super::{
     dbtsingleton_private::DbtSingletonPrivate, traits::*, RustyKey, RustyValue, WriteOperation,
 };
-use crate::sync::AtomicRw;
+use crate::sync::{AtomicRw, LockCallbackFn};
 use serde::{de::DeserializeOwned, Serialize};
 
 /// Singleton type created by [`super::DbtSchema`]
@@ -40,7 +40,12 @@ where
 {
     // DbtSingleton can not be instantiated directly outside of this crate.
     #[inline]
-    pub(crate) fn new(key: RustyKey, reader: Arc<dyn StorageReader + Sync + Send>) -> Self {
+    pub(crate) fn new(
+        key: RustyKey,
+        lock_name: String,
+        reader: Arc<dyn StorageReader + Sync + Send>,
+        lock_callback_fn: Option<LockCallbackFn>,
+    ) -> Self {
         let singleton = DbtSingletonPrivate::<V> {
             current_value: Default::default(),
             old_value: Default::default(),
@@ -48,7 +53,7 @@ where
             reader,
         };
         Self {
-            inner: AtomicRw::from(singleton),
+            inner: AtomicRw::from((singleton, Some(lock_name), lock_callback_fn)),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


dan-da:
In the interest of saving time, I merged this without review.   The changes are all optional and backwards compatible, with exception of the one breaking change, discussed above.

PR #174, opened 2024-01-10
chore: add windows, mac to github CI tasks

dan-da:


dan-da:
hmm, I see that this doesn't display checks for each OS in the `checks` list (and it may not work at all).  I will revise it to use a runner-matrix as in neptune-core.

dan-da:
Ok, well the checks all completed.    Linux and Mac passed.   Windows failed test `storage::level_db::tests::level_db_close_and_reload`.   I will file an issue for that and see about fixing it. 

Anyway, the checks seems to work, so I am merging this.

PR #176, opened 2024-01-10
fix: remove DB file on Drop, for windows.

### LINKED ISSUE #175: test storage::level_db::tests::level_db_close_and_reload fails on windows.
CI reports this failure on windows only:

```
---- storage::level_db::tests::level_db_close_and_reload stdout ----
thread 'storage::level_db::tests::level_db_close_and_reload' panicked at twenty-first\src\storage\level_db.rs:310:9:
assertion failed: !db_path.exists()

failures:
    storage::level_db::tests::level_db_close_and_reload
```

dan-da:
fixes #175

make `storage::level_db::tests::level_db_close_and_reload` pass on windows.

<strike>we create a temp db in order to drop the existing DB, which is preventing deletion of the underlying file.</strike>

We make `DB::db` an `Option` which allows gaining ownership of the `leveldb::Database` inside `impl Drop`.  We can then drop the `Database`, which closes the underlying file and allows deletion.

<strike>This is a draft PR until I verify tests pass  on windows.</strike>

Tests are passing on all 3 platforms.

### DIFF
--- a/twenty-first/src/storage/level_db.rs
+++ b/twenty-first/src/storage/level_db.rs
@@ -16,6 +16,8 @@ use leveldb::{
     options::{Options, ReadOptions, WriteOptions},
     snapshots::{Snapshot, Snapshots},
 };
+use rand::distributions::DistString;
+use rand_distr::Alphanumeric;
 use std::path::Path;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


dan-da:
this fix passes tests on windows, which proves the theory/cause, however it still leaves around a lot of tmp files when running tests.   I have another fix in mind that will not do that.   incoming later today.

dan-da:
I made a better fix that does not require any new tmp file.    Merging...

PR #179, opened 2024-01-11
Fix potential db-key collission for singleton and dbtvec length

### LINKED ISSUE #177: Add Key Prefix to DbtSingleton to avoid collisions.
In issue #163, @Sword-Smith commented:

> 1. I notice that `DbtSingletonPrivate` doesn't have a `key_prefix` value like `DbtVecPrivate` has. Shouldn't it have that to guarantee that key-collissions cannot happen?

I'm opening this issue for discussion.

Sword-Smith:
By adding a deterministically derived value from the schema, the number of tables added to the schema, and using this as a key for the underlying leveldb database, we can ensure that no key collissions happen, as all tables have a unique prefix that *all* their keys start with.

This closes #177.

### DIFF
--- a/twenty-first/src/storage/storage_schema/dbtsingleton.rs
+++ b/twenty-first/src/storage/storage_schema/dbtsingleton.rs
@@ -1,8 +1,6 @@
 use std::{fmt::Debug, sync::Arc};
 
-use super::{
-    dbtsingleton_private::DbtSingletonPrivate, traits::*, RustyKey, RustyValue, WriteOperation,
-};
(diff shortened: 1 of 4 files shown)


dan-da (changes requested):
LGTM, though I'd like to change the name param so we don't clone if we don't have to.  Definitely not critical here, but I think `impl Into<String>` is a good pattern to adopt in general.

also, it needs a rebase.

Sword-Smith:
Implemented all your changes, rebased, and merged.

PR #181, opened 2024-01-21
Add storage::DbtMap, for key/val within a DbSchema.

dan-da:
note: This draft PR is based on branch `0_36_backports_mut_self_squashed` which is waiting to merge to master until triton-vm+tasm-lib+neptune-core are building with master again.

This adds a DbtMap type.  It is an experiment, as really we are creating a logical key/val subset within the full key/val set of LevelDB, and LevelDB provides no facilities for that, so it is awkward.   With this initial design/impl, DbtMap should only be used for small sets of values, probably less than 1000 entries.    Here is the comment from DbtMapInternal, with more detail:

```
Important!  This type should only be used for storing
a small number of values, eg under 1000, although there
is no fixed limit.

DbtMap is essentially creating a logical sub-set of key/val
pairs inside the full set, which is the LevelDB database.

In order to persist its list of known keys, DbtMap stores
that list in a single LevelDB key, which must be updated for
every insert or delete.

Additionally, DbtMap keeps a copy of the key-list in RAM,
which is loaded from the DB during initialization.

For this reason, this type is not suited to storing
huge numbers of entries.

So in this design, for a DbtMap with prefix `0` and values
`{(10,100), (20,200), (30,300)}`, the levelDB representation logically
looks like:
  0_kl: [10,20,30],
  0_10: 100,
  0_20: 200,
  0_30: 300

So the entire `0_kl` list must be updated for each insertion/removal.

An alternative design is possible where the DbtMap
keys are stored in an ascending list of LevelDB keys,
eg:
  0_kl_0: 10,
  0_kl_1: 20,
  0_kl_2: 30,
  0_10: 100,
  0_20: 200,
  0_30: 300

Here only a single small `0_kl_x` key needs to be added or removed
for each insertion/removal.

This design could scale better to large numbers of entries
but would be slower for read access without all keys
cached in RAM.
```

Really though, *both* designs are awkward as every insertion/removal requires two writes instead of one, and we require (up-to) double the space in the DB in order to store the key list.  But without a key-list, the only way to find "our" keys would be to iterate over the entire LevelDB, which doesn't seem a good option either.

So basically, this is not a good fit for LevelDB.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,89 +9,16 @@ on:
 name: Rust
 
 jobs:
-  check:
-    name: Check
-    runs-on: ubuntu-latest
+  runner-matrix:
+    name: format, lint, test
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout sources
-        uses: actions/checkout@e2f20e631ae6d7dd3b768f56a5d2af784dd54791 # v2.5.0, 2022-10-13
-
-      - name: Install stable toolchain
-        uses: actions-rs/toolchain@b2417cde72dcf67f306c0ae8e0828a81bf0b189f # v1.0.7, 2020-03-24
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-
-      - name: Run cargo check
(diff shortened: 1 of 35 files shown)


dan-da:
Closing this as:

1. storage has been move out of twenty-first and into neptune_core::database::storage.
2. this DbtMap is too slow/inefficient for large datasets to be useful.

PR #182, opened 2024-01-22
evict interior mutability in storage.  (&self --> &mut self)

dan-da:
<strike>This is a draft PR until triton-vm and tasm-lib are made compatible with twenty-first (master).   At that time, I will rebase this on master.</strike>.   rebased.  ready for review.

---

The basic premise is to have methods take `&mut self` instead of `&self`, most everywhere (in `storage` and also `neptune-core`).

The rationale is that `&self` hides where mutations are occurring, and this propagates up the call-chain.  So it can look like a fn is immutable, but really it is mutating something 3 calls down.

This does not change any functionality or behavior.   But it does change the public API.

https://github.com/Neptune-Crypto/neptune-core/pull/91 <strike>depends on</strike> complements this.

### DIFF
--- a/twenty-first/benches/db_dbtvec.rs
+++ b/twenty-first/benches/db_dbtvec.rs
@@ -101,7 +101,7 @@ mod write_100_entries {
         use super::*;
 
         fn push_impl(bencher: Bencher, persist: bool) {
-            let (mut storage, vector) = create_test_dbtvec();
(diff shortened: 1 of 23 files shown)


dan-da:
closing.   In the interest of saving time, I just pushed to master without review.    Let me know if any concern.

PR #184, opened 2024-01-27
Allow repeated sponge-absorption without mutating a list

### LINKED ISSUE #183: Add function to `SpongeHasher` that allows repeated abosorption without mutating a list
From the Virtual machine we want to call `absorb_repeatedly` on data from non-deterministically initialized memory. But the current interface requires the caller to handle the padding which would require a reallocation of the entire list found in memory. To avoid this waste of clock cycles, we should add a function that can be called on a slice and that handles padding for us.

Sword-Smith:
Also renames two `SpongeHasher` trait functions.

This closes #183.

### DIFF
--- a/twenty-first/src/shared_math/tip5.rs
+++ b/twenty-first/src/shared_math/tip5.rs
@@ -921,7 +921,7 @@ mod tip5_tests {
             let hash_varlen_digest = Tip5::hash_varlen(&preimage);
 
             let mut sponge = Tip5::init();
-            Tip5::pad_and_absorb_repeatedly(&mut sponge, &preimage);
+            Tip5::pad_and_absorb_all(&mut sponge, &preimage);
             let squeeze_result = Tip5::squeeze_once(&mut sponge);
             let digest_through_pad_squeeze_absorb =
                 Digest::new((&squeeze_result[..DIGEST_LENGTH]).try_into().unwrap());
@@ -936,7 +936,7 @@ mod tip5_tests {
         let hash_varlen_digest = Tip5::hash_varlen(&preimage);
 
         let mut sponge = Tip5::init();
-        Tip5::pad_and_absorb_repeatedly(&mut sponge, &preimage);
+        Tip5::pad_and_absorb_all(&mut sponge, &preimage);
         let squeeze_result = Tip5::squeeze_once(&mut sponge);
         let digest_through_pad_squeeze_absorb =
             Digest::new((&squeeze_result[..DIGEST_LENGTH]).try_into().unwrap());
--- a/twenty-first/src/shared_math/tip5.rs
+++ b/twenty-first/src/shared_math/tip5.rs
@@ -3,16 +3,20 @@ use bfieldcodec_derive::BFieldCodec;
 use get_size::GetSize;
 use itertools::Itertools;
 use num_traits::Zero;
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
+use serde::Serialize;
 
-use crate::shared_math::b_field_element::{BFieldElement, BFIELD_ONE, BFIELD_ZERO};
-pub use crate::shared_math::digest::{Digest, DIGEST_LENGTH};
-
-use crate::util_types::algebraic_hasher::{AlgebraicHasher, Domain, SpongeHasher};
-
+use crate::shared_math::b_field_element::BFieldElement;
+use crate::shared_math::b_field_element::BFIELD_ONE;
+use crate::shared_math::b_field_element::BFIELD_ZERO;
+pub use crate::shared_math::digest::Digest;
+pub use crate::shared_math::digest::DIGEST_LENGTH;
 use crate::shared_math::mds::generated_function;
 use crate::shared_math::x_field_element::XFieldElement;
 use crate::shared_math::x_field_element::EXTENSION_DEGREE;
+use crate::util_types::algebraic_hasher::AlgebraicHasher;
+use crate::util_types::algebraic_hasher::Domain;
+use crate::util_types::algebraic_hasher::SpongeHasher;
 
 pub const STATE_SIZE: usize = 16;
 pub const NUM_SPLIT_AND_LOOKUP: usize = 4;
@@ -619,7 +623,7 @@ impl AlgebraicHasher for Tip5 {
         let num_squeezes = (num_elements + xfes_per_squeeze - 1) / xfes_per_squeeze;
         (0..num_squeezes)
             .flat_map(|_| {
-                Self::squeeze_once(state)
+                Self::squeeze(state)
                     .into_iter()
                     .take(xfes_per_squeeze * EXTENSION_DEGREE)
                     .collect_vec()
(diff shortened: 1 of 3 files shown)


Sword-Smith:
@aszepieniec Please double-check that the `SpongeHasher` trait has all we need, and that you approve of the function names.

Sword-Smith:
After merging this, next step should be to align snippet names in `tasm-lib` with the function names in the `SpongeHasher` and `AlgebraicHasher` traits.

twenty-first/src/util_types/algebraic_hasher.rs
```diff
@@ -55,9 +55,17 @@ pub trait SpongeHasher: Clone + Debug + Default + Send + Sync {
                 .collect::<Vec<_>>()
                 .try_into()
                 .expect("a multiple of RATE elements");
-            Self::absorb(sponge, &absorb_elems);
+            Self::absorb_once(sponge, &absorb_elems);
         }
     }
+
+    fn pad_and_absorb_repeatedly(sponge: &mut Self::SpongeState, input: &[BFieldElement]) {
```
aszepieniec, line 62: I prefer `pad_and_absorb_all` over `pad_and_absorb_repeatedly`. The reason is that "repeatedly" is confusing when only one iteration is necessary.

Sorry for the unstable recommendation.

aszepieniec (changes requested):
With the proposed name change, I am happy to sign off on this PR.

Before merging let's ask and answer this question: The sponge interface (as defined by the various sponge construction papers) defines "absorb" and "squeeze" and nothing else. This is what cryptographers are familiar with. (The papers also mention a padding rule to be applied prior to absorbing.) The question is why don't we follow this interface? More specifically: do we need the interface to distinguish between 1 and more than 1 iterations? (If we do, then the suffices "once" and "repeatedly" are well chosen.)

twenty-first/src/util_types/algebraic_hasher.rs
Sword-Smith: Should the name `absorb_repeatedly` also be change to `absorb_all` then?

aszepieniec: Nah, "repeatedly" is fine if it is the opposite of "once". For `pad_and_absorb_all` there is no "once" case.

Sword-Smith:
Now that I think about it, why don't we delete the function `absorb_repeatedly` entirely and only leave the version that pads? That way the padding logic doesn't have to be repeated multiple places.

aszepieniec:
> Now that I think about it, why don't we delete the function absorb_repeatedly entirely and only leave the version that pads? That way the padding logic doesn't have to be repeated multiple places.

I would prefer that but I hesitate because it might break something downstream ... Or introduce a vulnerability without causing a test to fail, as a result of violating now-incorrect assumptions about how the interface is supposed to work.

Likewise, I would prefer to drop the distinction between "once" and "repeatedly". Ideally:
 - `absorb` takes an `Iterator` and when the iterator reaches its end, the function applies the padding.
 - `squeeze` gives an `Iterator`.

Sword-Smith:
> I would prefer that but I hesitate because it might break something downstream

I already fixed the downstream of `triton-vm`, `tasm-lib`, and `tasm-lang` with the old `absorb_repeatedly` function removed. Those changes live on `sponge-hasher-trait` branches. Those changes simply consisted of removing manual padding, e.g. removing repeated logic, so I see that as a benefit.

> Or introduce a vulnerability without causing a test to fail, as a result of violating now-incorrect assumptions about how the interface is supposed to work.
We have very good test coverage. Trust the tests.

> Likewise, I would prefer to drop the distinction between "once" and "repeatedly". Ideally:
One of the purposes of this PR is to give clear and efficient counterparts that we can run in Triton-VM, i.e. to align host-machine code and Triton-VM code. with the currently suggested interface, I believe I can do that. With your suggested interface with iterators, that requires *a lot* more work that I don't see bearing any significant benefit.

The current suggestion is:
```rust
/// Initialize a sponge state
fn init() -> Self::SpongeState

/// Absorb an array of [RATE] field elements into the sponge's state, mutating it.
fn absorb_once(sponge: &mut Self::SpongeState, input: &[BFieldElement; RATE])

/// Squeeze an array of [RATE] field elements out from the sponge's state, mutating it.
fn squeeze_once(sponge: &mut Self::SpongeState) -> [BFieldElement; RATE]

/// Chunk `input` into arrays of [RATE] elements and repeatedly [SpongeHasher::absorb()].
fn pad_and_absorb_all(sponge: &mut Self::SpongeState, input: &[BFieldElement])
```

aszepieniec:
```language=rust
/// Absorb an array of [RATE] field elements into the sponge's state, mutating it.
fn absorb_once(sponge: &mut Self::SpongeState, input: &[BFieldElement; RATE]);
```

Does `absorb_once` apply padding? If yes, then `hash_pair` works without invoking this interface, which seems odd but not necessarily bad. Either way, the docstring should clearly mention whether padding is applied and warn against misuse.

aszepieniec (approved):

twenty-first/src/shared_math/tip5.rs
```diff
@@ -909,6 +912,37 @@ mod tip5_tests {
         );
     }
 
+    #[test]
+    fn hash_var_len_equivalence_corner_cases() {
+        // Verify that `hash_var_len` is equivalent
+        // to pad + repeated absorb + squeeze once
+        for preimage_length in [0, 1, 2, 3, 4, 5, 6, 9, 10, 11] {
```
jan-ferdinand, line 919: might as well do
```suggestion
        for preimage_length in 0..=11 {
```
or
```suggestion
        for preimage_length in (0..=6).chain(9..=11) {
```
if leaving out 7 and 8 is important. The long list of integers is time consuming to parse imo.

twenty-first/src/shared_math/tip5.rs
```diff
@@ -909,6 +912,37 @@ mod tip5_tests {
...
+            let preimage = random_elements(preimage_length);
```
jan-ferdinand, line 920: Let's make this a `#[proptest]`, too. In my opinion, we should never use system randomness directly in tests. Randomness should alway be controlled by the testing framework.

twenty-first/src/shared_math/tip5.rs
```diff
@@ -909,6 +912,37 @@ mod tip5_tests {
...
+            let hash_varlen_digest = Tip5::hash_varlen(&preimage);
+
+            let mut sponge = Tip5::init();
+            Tip5::pad_and_absorb_all(&mut sponge, &preimage);
+            let squeeze_result = Tip5::squeeze_once(&mut sponge);
+            let digest_through_pad_squeeze_absorb =
+                Digest::new((&squeeze_result[..DIGEST_LENGTH]).try_into().unwrap());
+            assert_eq!(digest_through_pad_squeeze_absorb, hash_varlen_digest);
```
jan-ferdinand, line 928: These lines shared identically in this test and the test below it. Let's factor them out into a method.

jan-ferdinand:
While we're in the process of touching interfaces, how about signatures such as the following:

```rust
pub trait SpongeHasher: Clone + Debug + Default + Send + Sync {
    const RATE: usize;
    type SpongeState: Clone + Debug + Default + for<'a> Arbitrary<'a>;

    fn init() -> Self;
    fn absorb(&mut self, input: &[BFieldElement; RATE]);
    fn squeeze(&mut self) -> [BFieldElement; RATE];
    fn pad_and_absorb_all(&mut self, input: &[BFieldElement]);
}
```

PR #185, opened 2024-02-09
simplify `SpongeHasher` trait

jan-ferdinand:
This is an implementation of the suggestion made [in this comment](https://github.com/Neptune-Crypto/twenty-first/pull/184#issuecomment-1914312403) on #184. Concretely, I propose to use `fn foo(&mut self, …)` in favor of `fn foo(the_thing: &mut TheThing, …)` wherever possible. This allows to write, for example, `self.squeeze()` as opposed to `Self::squeeze(&mut sponge_state)`.

Additionally, I propose to merge the unit struct `Tip5` with the state struct `Tip5State` under the rationale that they are very tightly coupled and we gain no benefit from pulling the two apart.[^blake]

Of note and probably of particular interest to @aszepieniec: I was unable to detect any performance change in `Tip5` across the 3 benchmarks I ran.

Other, minor points:

- rename `SpongeHasher` to `Sponge`
- improve documentation (includes deleting statements that always were or are now obvious from a function's signature)

[^blake]: This also holds for `Blake3` and `Blake3State`. However, I understand this to be legacy code and therefore of lesser interest.

### DIFF
--- a/twenty-first/src/prelude.rs
+++ b/twenty-first/src/prelude.rs
@@ -11,7 +11,7 @@ pub use crate::shared_math::traits::ModPowU32;
 pub use crate::shared_math::x_field_element;
 pub use crate::shared_math::x_field_element::XFieldElement;
 pub use crate::util_types::algebraic_hasher::AlgebraicHasher;
-pub use crate::util_types::algebraic_hasher::SpongeHasher;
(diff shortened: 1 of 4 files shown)


aszepieniec (approved):
Much cleaner. Well done. :clapper:

PR #188, opened 2024-02-27
simplify field element construction

jan-ferdinand:
The main addition are the `bfe!` and `xfe!` macros.  They simplify construction of `BFieldElements` and `XFieldElements`, respectively.

Additionally,
- `BFieldElement::from` can now deal with `u8`, `u16`, and `i32`.
- `XFieldElement::from` can now deal with `T` and `[T; EXTENSION_DEGREE]` where `T` is any type that can be converted into a `BFieldElement`.

### DIFF
--- a/twenty-first/src/prelude.rs
+++ b/twenty-first/src/prelude.rs
@@ -1,4 +1,5 @@
 pub use crate::amount::u32s::U32s;
+pub use crate::bfe;
 pub use crate::shared_math::b_field_element;
 pub use crate::shared_math::b_field_element::BFieldElement;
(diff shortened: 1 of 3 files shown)


Sword-Smith (commented):
Looks like a really nice addition.

I would add a test of the macros `xfe!` and `bfe!` that they actually produce the expected values. Something like `assert_eq!(BFieldElement::new(14), bfe!(14));`

Sword-Smith (approved):
LGTM

aszepieniec (approved):
LGTM

PR #190, opened 2024-03-14
replace `ModPowU32` etc with `Pow<T>`

jan-ferdinand:
I prefer one generic `Pow<T>` trait over two non-generic `ModPowU32` and `ModPowU64` traits. This PR implements this preference using `num_traits::Pow`.


Sword-Smith:
The `mod_pow_u32` is implemented with these names in both `tasm-lib` and `tasm-lang`. So in `tasm-lang`, you can e.g. `xfe.mod_pow_u32(4)`.  This PR would make the repos drift apart, unless this change is carried downstream.

jan-ferdinand:
Sad 🤷

PR #192, opened 2024-03-23
remove cruft

jan-ferdinand:
Take the chainsaw to the codebase. ⛓🪚

## Open questions

- [x] `fips202.rs` – while SHA-3 is used in `lattice.rs`, do we want to roll it ourselves? [SHA-3 is popular](https://crates.io/crates/sha3).
- [x] `dyadic_rationals.rs`?
- [x] `trait FromVecu8`?
- [x] `trait Emojihash`?
- [x] rename `shared_math` into `math`?
- [x] `trait New` defining `new_from_usize`

What do you (@aszepieniec, @Sword-Smith, @dan-da) think we can and should remove?


Sword-Smith:
EmojiHash has been a stone in my shoe for some time. Happy to see that one gone :)

jan-ferdinand:
Let me know if I went overboard with anything.

aszepieniec (approved):

twenty-first/src/util_types/mmr/shared_advanced.rs
```diff
@@ -144,11 +142,35 @@ pub fn get_authentication_path_node_indices(
...
+/// ```
+/// # use twenty_first::util_types::mmr::shared_advanced::get_peak_heights;
+/// assert_eq!(get_peak_heights(0b1010), vec![3, 1]);
+/// assert_eq!(get_peak_heights(0b1011), vec![3, 1, 0]);
+/// ```
 pub fn get_peak_heights(leaf_count: u64) -> Vec<u8> {
-    // The peak heights in an MMR can be read directly from the bit-decomposition
-    // of the leaf count.
-    indices_of_set_bits(leaf_count)
+    // In an MMR, the peak heights directly correspond the leaf count's bit decomposition. That is,
```
Sword-Smith, line 155: Eht this rewrite? Otherwise this PR looks good.

jan-ferdinand: Main reason because the used method was only ever used once, across all our crates. This made the corresponding MMR method very shallow. Citing [46531c4](https://github.com/Neptune-Crypto/twenty-first/pull/192/commits/46531c4a7b6071856ae0937382887c09a04e57eb):

> Function `indices_of_set_bits` seemed like it could be useful as a general function. However, in practice, it is only used in Merkle Mountain Ranges.

That said, I'm not married to this change. :relieved:

Sword-Smith (approved):

PR #193, opened 2024-04-15
Thv/fix scale 1

Sword-Smith:
Fix performance degredation of `scale` and add benchmarks to run when a new version is published. 

Breaking change, as the `scale` method is renamed.


jan-ferdinand:
I think all the changes made here have been incorporated into #194. Is that assumption correct?

PR #194, opened 2024-04-15
Speed up various polynomial functions

jan-ferdinand:
Mostly engineering, sometimes math, to speed up performance-critical polynomial functions, such as `evaluate`, `interpolate`, `multiply`, etc. Canonical benchmarking target is `mjolnir`.


Sword-Smith (approved):
Looks good. One performance-related questions though:
`scalar_mul_mut` looks faster to me in some situations, as it avoids a re-allocation of a polynomial. Are you sure it's not faster than `scalar_mul` that you're replacing it with?

jan-ferdinand:
There appears to be a measurable difference between the two methods:

```
Multiplication of Polynomial of Degree 2^1 with a Scalar/Mut/1
                        time:   [4.1544 ns 4.1862 ns 4.2327 ns]
Multiplication of Polynomial of Degree 2^1 with a Scalar/Immut/1
                        time:   [17.486 ns 17.583 ns 17.725 ns]

Multiplication of Polynomial of Degree 2^7 with a Scalar/Mut/7
                        time:   [135.95 ns 136.72 ns 137.65 ns]
Multiplication of Polynomial of Degree 2^7 with a Scalar/Immut/7
                        time:   [142.00 ns 142.95 ns 144.27 ns]

Multiplication of Polynomial of Degree 2^13 with a Scalar/Mut/13
                        time:   [8.4872 µs 8.5008 µs 8.5170 µs]
Multiplication of Polynomial of Degree 2^13 with a Scalar/Immut/13
                        time:   [10.081 µs 10.142 µs 10.221 µs]
```

I've removed the deprecation, added a benchmark highlighting the performance difference, changed all callers (for which it makes sense) to use the mutable variant, and added documentation.

Interestingly, as far as I can tell, we have not used the method for a while.

Issue #195, opened 2024-04-19
bug: digest bytes in do not match bytes out.


dan-da:
In the course of testing hex encode/decode, I discovered that it is possible to create a `Digest` from bytes that subsequently produces different bytes.

Here is a failing test case, suitable for use in digest.rs.

```
    #[test]
    pub fn bytes_in_matches_bytes_out() {
        let bytes1: [u8; 40] = [255; 40];
        let d1 = Digest::from(bytes1);

        let bytes2 = <[u8; 40]>::from(d1);
        let d2 = Digest::from(bytes2);

        println!("bytes1: {:?}", bytes1);
        println!("bytes2: {:?}", bytes2);

        assert_eq!(d1, d2);
        assert_eq!(bytes1, bytes2);
    }
```

This test fails with:

```
---- math::digest::digest_tests::bytes_in_matches_bytes_out stdout ----
bytes1: [255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255]
bytes2: [254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0]
thread 'math::digest::digest_tests::bytes_in_matches_bytes_out' panicked at twenty-first/src/math/digest.rs:474:9:
assertion `left == right` failed
  left: [255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255]
 right: [254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0, 254, 255, 255, 255, 0, 0, 0, 0]
```

note that:
1. `bytes2` is quite different from `bytes1`.
2. digests `d1` and `d2` are equal, although d2 was created from bytes2.

[...]

PR #196, opened 2024-04-19
feat: add hex encoding and decoding for Digest

dan-da:
Adds Digest::to_hex() and Digest::try_from_hex().

Also:
 * Adds a dep on `hex` crate.
 * Adds FromHexDigestError error enum
 * Adds TryFrom<&[u8]> for Digest, which is used by ::try_from_hex()
 * Adds tests for hex encoding and decoding
 * Adds tests for bytes --> Digest --> bytes roundtrip.

**Note that some hex tests and the bytes roundtrip tests are presently failing due to issue #195, which needs to be resolved.**

The hex functions will be useful for a block explorer and probably many other use-cases down the road.  It is not strictly necessary to have these fn inside twenty-first, but makes life easier for api users and provides a consistent way to encode/decode Digests for end-users to see throughout neptune ecosystem.

Also the `TryFrom<&[u8]>` seems generally useful.  While it doesn't have dedicated test(s), it has test coverage as it is used by ::try_from_hex().


dan-da:
related: https://bitcoin.stackexchange.com/questions/122733/why-doesnt-bitcoin-use-bech32-encoding-for-transaction-id-and-block-hash/122734#122734

twenty-first/src/math/digest.rs
```diff
@@ -416,4 +472,205 @@ pub(crate) mod digest_tests {
...
+            let hex_vals = vec![
+                "00000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                "10000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                "0000000000000000000000000000000000000000000000000000000000000000000000000000000f",
+                "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+                // note: this would cause test failures due to:
+                //       https://github.com/Neptune-Crypto/twenty-first/issues/195
+                // "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+            ];
+            for hex in hex_vals.into_iter() {
```
jan-ferdinand, line 583:
```suggestion
            for hex in hex_vals {
```

twenty-first/src/math/digest.rs
```diff
@@ -416,4 +472,205 @@ pub(crate) mod digest_tests {
...
+
+    mod hex_test {
+
+        use super::*;
+
+        pub(super) fn hex_examples() -> Vec<(Digest, &'static str)> {
+            vec![
+                (
+                    Digest::default(),
+                    "00000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
jan-ferdinand, line 538: nit: This is longer than the maximum line width, which is 100 characters. Perhaps a different arrangement reduces the indentation level?

twenty-first/src/math/digest.rs
```diff
@@ -416,4 +472,205 @@ pub(crate) mod digest_tests {
...
+            ];
+            for hex in hex_vals.into_iter() {
+                let digest = Digest::try_from_hex(hex).unwrap();
+                assert_eq!(hex, &digest.to_hex())
+            }
+        }
+
+        #[test]
+        pub fn digest_from_hex() -> Result<(), FromHexDigestError> {
+            for (digest, hex) in hex_examples().into_iter() {
```
jan-ferdinand, line 591:
```suggestion
            for (digest, hex) in hex_examples() {
```

twenty-first/src/math/digest.rs
```diff
@@ -416,4 +472,205 @@ pub(crate) mod digest_tests {
...
     }
+
+    // note: this test panics due to:
+    //       https://github.com/Neptune-Crypto/twenty-first/issues/195
+    //
+    //       for now we use a #[should_panic]. Depending on resolution
+    //       of #195, we may be able to remove it.
+    #[test]
+    #[should_panic]
+    pub fn bytes_in_matches_bytes_out() {
```
jan-ferdinand, line 483: nit: Is there a reason the test is `pub`? Also applies to the other tests.

twenty-first/src/math/digest.rs
jan-ferdinand: And so on below

twenty-first/src/math/digest.rs
```diff
@@ -206,6 +217,51 @@ impl Digest {
...
     }
+
+    /// Encode digest as hex
+    pub fn to_hex(&self) -> String {
+        let bytes = <[u8; Digest::BYTES]>::from(*self);
+        hex::encode(bytes)
+    }
+
+    /// Decode hex string to Digest
+    pub fn try_from_hex(string: &str) -> Result<Self, FromHexDigestError> {
```
jan-ferdinand, line 228: Would it make sense to use the same trait bound as `hex::decode`?
```suggestion
    pub fn try_from_hex(data: impl AsRef<[u8]>) -> Result<Self, FromHexDigestError> {
```

jan-ferdinand (changes requested):
I like this. I think it's beneficial to clear up #195 before releasing some next version that contains this PR; however, I'm fine merging this and addressing #195 separately.

I've left some comments in-line.

twenty-first/src/math/digest.rs
```diff
@@ -11,18 +11,19 @@ use rand::Rng;
 use rand_distr::Distribution;
 use rand_distr::Standard;
 use serde::Deserialize;
+use serde::Deserializer;
 use serde::Serialize;
+use serde::Serializer;
 
-use crate::error::TryFromDigestError;
+use crate::error::{FromHexDigestError, TryFromDigestError};
```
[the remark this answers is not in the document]
dan-da: yeah I've been drifting towards your position over time.   The primary reason is that the single-value-per-line is much easier to work with when resolving git conflicts.

I think the counterpoints would be that single-value-per-line is more verbose and may obscure (at a glance) that types come from the same crate or module.

That said, for my workflow, the git-conflict (diff) thing outweighs the counterpoints, so I'm happy to make the change and use that style going forward.

At present, we use a mixed style, at least in neptune-core.  It would probably be good to settle on a style across neptune-crates.   I wonder if there is tooling support for this, eg a `cargo fmt` flag, clippy lint, etc.   Ideally, `cargo fmt` or similar tool would just reformat each crate for us...

twenty-first/src/math/digest.rs
```diff
@@ -206,6 +217,51 @@ impl Digest {
     pub fn hash<H: AlgebraicHasher>(self) -> Digest {
         H::hash_pair(self, Digest::new([BFieldElement::zero(); DIGEST_LENGTH]))
     }
+
+    /// Encode digest as hex
+    pub fn to_hex(&self) -> String {
+        let bytes = <[u8; Digest::BYTES]>::from(*self);
```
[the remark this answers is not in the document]
dan-da: good point.  Just force of habit.  I'll make the change.

twenty-first/src/math/digest.rs
```diff
@@ -206,6 +217,51 @@ impl Digest {
...
+    /// Decode hex string to Digest
+    pub fn try_from_hex(string: &str) -> Result<Self, FromHexDigestError> {
+        let slice = hex::decode(string)?;
+        Ok(Self::try_from(&slice as &[u8])?)
+    }
+}
+
+// we implement Serialize so that we can serialize as
+// hex for human readable formats like JSON but use
+// default serializer for other formats likes bincode
```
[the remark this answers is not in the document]
dan-da: yeah, that one is a bit short.  I will reflow it to 80.

Coming from C way back when I've always kinda stuck with 80 cols as a soft-limit for both comments and code.  But moreso for comments.

I see that that rustfmt uses [80 for comments](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#comment_width) and 100 for code.  The style guide [agrees](https://doc.rust-lang.org/beta/style-guide/index.html#comments).

twenty-first/src/math/digest.rs
```diff
@@ -206,6 +217,51 @@ impl Digest {
...
+impl Serialize for Digest {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        if serializer.is_human_readable() {
```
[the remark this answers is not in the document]
dan-da: yeah I was very happy to find it.

twenty-first/src/math/digest.rs
dan-da: this has been through `cargo fmt`.  I'm open to suggestions.

possibly a const vec instead of the hex_examples() fn.   I'll play with it.

twenty-first/src/math/digest.rs
dan-da: no, just habit.  That seems to be the convention in neptune-core.  I notice that Digest::test::digest_from_str() is also pub.  I will change them all to be private, for consistency.

twenty-first/src/math/digest.rs
dan-da: yes!  good catch, thx.   I meant to go back and do that, and it slipped my mind.  Friendlier for callers.

twenty-first/src/math/digest.rs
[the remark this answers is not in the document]
jan-ferdinand: Actually, I concur with the git conflicts being the primary reason to avoid this. Seems like I have forgotten how annoying that is. (I'm using single-line style in the crates I mainly work on for a while now.)

I'd be stoked to learn about a tool that automates this conversion. Currently, I basically run `grep "^ *use.*\{"` to see if any violations have snuck in, which is admittedly a terrible workflow.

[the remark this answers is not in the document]
jan-ferdinand: From what I can tell, [this](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#imports_granularity) is the relevant option on `rustfmt`. There is an open tracking issue with recent activity linked there.

twenty-first/src/math/digest.rs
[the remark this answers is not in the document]
jan-ferdinand: Interesting. Thanks for the resource. :+1:

twenty-first/src/math/digest.rs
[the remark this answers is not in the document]
jan-ferdinand: >  It would probably be good to settle on a style across neptune-crates

I also concur with this. As far as I recall, @Sword-Smith is also in favor of the single-line style – please correct me if I misremember. That said, it's very tedious to manually change everything. :face_exhaling:

twenty-first/src/math/digest.rs
dan-da: I addressed this with:

```
                    concat!(
                        "0000000000000000000000000000000000000000",
                        "0000000000000000000000000000000000000000"
                    ),
```

jan-ferdinand (approved):

twenty-first/src/math/digest.rs
jan-ferdinand: Even though it requires naming more things, I find the following arrangement visually pleasing, too. But we're definitely in the “kinda whatever :shrug:” area now. :relieved: 

```rust
pub(super) fn hex_examples() -> Vec<(Digest, &'static str)> {
    let example_0 = (
        Digest::default(),
        "0000000000000000000000000000000000000000\
         0000000000000000000000000000000000000000",
    );
    let example_1 = /* … */;

    // note: this would cause test failure due to:
    //       https://github.com/Neptune-Crypto/twenty-first/issues/195
    let _example_2 = (
        Digest::new(bfe_array![0, 1, 10, 15, 255]),
        "ffffffffffffffffffffffffffffffffffffffff\
         ffffffffffffffffffffffffffffffffffffffff",
    );

    vec![example_0, example_1]
}
```

PR #199, opened 2024-04-23
fix: Digest::try_from() returns NotCanonical error

dan-da:
Addresses #195.

Any attempt to instantiate a Digest from bytes or a string with embedded non-canonical u64 values now results in an error.

Note that it is still possible to instantiate a Digest from BFieldElements that were created from non-canonical u64. Maybe that's ok...?

Changes:
 * adds TryFromDigestError::NotCanonical
 * adds BFieldElement::is_canonical()
 * Digest::from_str() returns NotCanonical error
 * Digest::try_from(bytes) returns NotCanonical error
 * adds Digest test try_from_bytes_not_canonical()
 * adds Digest test from_str_not_canonical()
 * adds Canonical test in digest_from_invalid_hex_errors()

--------

Commentary:

I am not entirely happy with this solution/pr for two reasons:

1. it only partially addresses the issue. It remains possible to bypass the canonical checks by instantiating Digest from BFieldElements directly.  That feels wrong to me (leaky abstraction), but maybe it's actually Ok.   opinions welcome.
2. It seems cleaner to perform the canonical check(s) inside BFieldElement rather than Digest methods.

regarding (2) I initially started with this approach, and impl'd TryFrom bytes for BFieldElement.   This required getting rid of `impl From bytes for BFieldElement`.  That was fine as it doesn't seem to be used.  However, the `TryFrom bytes` is only a partial solution as BFieldElement still has a bunch of `impl From` for u64, u128, i32 etc as well as `impl FromStr`.   I felt I was likely to cause downstream breakage changing all these to `TryFrom` and also it was becoming a bigger project than anticipated, so I backed off.   I still feel that is the most correct fix if we really want to enforce the canonical check.   I pushed an [alternative branch](https://github.com/dan-da/twenty-first/tree/195_bfe_not_canonical_pr_alt) with this approach.

However, for basic bytes/string serialization, the present PR suffices, is smaller, and I verified that downstream neptune crates have no build errors.  I believe this is also closer to what was discussed/proposed in #125.

Another consideration is that its possible we might be instantiating Digest(s) from non-canonical BFieldElement inputs somewhere in the neptune stack.  So adding a check inside BFieldElement would have potential to break some logic (not just syntax).  I'm unsure how likely this is or not.

At this point we could merge this PR as-is, or pursue the other branch/approach.  I don't have a strong preference and am open to suggestions/feedback.

a final note:  Digest also has a TryFrom BigUint.  I looked at it, but it wasn't immediately clear to me if a canonical check is needed or exactly how to do it, so I didn't touch it.  That might be another leak.

### DIFF
--- a/twenty-first/src/error.rs
+++ b/twenty-first/src/error.rs
@@ -4,6 +4,7 @@ use thiserror::Error;
 pub use crate::math::bfield_codec::BFieldCodecError;
 use crate::prelude::tip5::DIGEST_LENGTH;
 use crate::prelude::x_field_element::EXTENSION_DEGREE;
+use crate::prelude::BFieldElement;
 pub use crate::util_types::merkle_tree::MerkleTreeError;
 
 #[derive(Debug, Clone, Eq, PartialEq, Error)]
@@ -39,7 +40,7 @@ pub enum TryFromDigestError {
     #[error("invalid `BFieldElement`")]
     InvalidBFieldElement(#[from] ParseBFieldElementError),
 
-    #[error("Not Canonical. ({0})")]
+    #[error("non-canonical {0} >= {} == `BFieldElement::P`", BFieldElement::P)]
     NotCanonical(u64),
 
     #[error("overflow converting to Digest")]
--- a/twenty-first/src/error.rs
+++ b/twenty-first/src/error.rs
@@ -39,13 +39,16 @@ pub enum TryFromDigestError {
     #[error("invalid `BFieldElement`")]
     InvalidBFieldElement(#[from] ParseBFieldElementError),
 
+    #[error("Not Canonical. ({0})")]
+    NotCanonical(u64),
+
     #[error("overflow converting to Digest")]
     Overflow,
 }
 
 #[derive(Debug, Clone, PartialEq, Error)]
 #[non_exhaustive]
-pub enum FromHexDigestError {
+pub enum TryFromHexDigestError {
     #[error("hex decoding error")]
     HexDecode(#[from] hex::FromHexError),
 
--- a/twenty-first/src/math/digest.rs
+++ b/twenty-first/src/math/digest.rs
@@ -104,13 +104,14 @@ impl FromStr for Digest {
         let maybe_parsed_u64s: Result<Vec<_>, _> =
             string.split(',').map(str::parse::<u64>).collect();
         let parsed_u64s = maybe_parsed_u64s.map_err(ParseBFieldElementError::ParseU64Error)?;
-        let mut bfes = vec![];
-        for val in parsed_u64s {
-            if !BFieldElement::is_canonical(val) {
-                return Err(TryFromDigestError::NotCanonical(val));
-            }
-            bfes.push(BFieldElement::from(val))
-        }
+
+        // checks if each u64 is canonical before instantiating into BFE.
+        let bfe_try_from = |v: u64| -> Result<BFieldElement, _> {
+            let bfe = BFieldElement::is_canonical(v).then(|| BFieldElement::new(v));
+            bfe.ok_or(TryFromDigestError::NotCanonical(v))
+        };
+        let bfes: Vec<_> = parsed_u64s.into_iter().map(bfe_try_from).try_collect()?;
+
         let invalid_len_err = Self::Err::InvalidLength(bfes.len());
         let digest_innards = bfes.try_into().map_err(|_| invalid_len_err)?;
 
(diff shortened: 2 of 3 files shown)


jan-ferdinand:
> Note that it is still possible to instantiate a Digest from BFieldElements that were created from non-canonical u64. Maybe that's ok...?

It's certainly okay if one has the `BFieldElement`s themselves, as they take care of what appears to be duplicates in `u64`-land. It can start being weird if one were to use `BFieldElement::from(my_u64)`. However, I think it's fine because a) this route seems convoluted and b) if one starts using `BFieldElement`s it can be expected from them to understand what they are. In contrast, having to understand `BFieldElement` only because one wants to get a `Digest` from a `&[u64]` seems wrong.

jan-ferdinand:
> Digest also has a TryFrom BigUint. I looked at it, but it wasn't immediately clear to me if a canonical check is needed or exactly how to do it, so I didn't touch it. That might be another leak.

I'm actually not sure whether we're using `BigUint` anywhere in our stack. Maybe @Sword-Smith or @aszepieniec happen to know?

twenty-first/src/math/digest.rs
```diff
@@ -100,11 +101,18 @@ impl FromStr for Digest {
...
+        let maybe_parsed_u64s: Result<Vec<_>, _> =
+            string.split(',').map(str::parse::<u64>).collect();
+        let parsed_u64s = maybe_parsed_u64s.map_err(ParseBFieldElementError::ParseU64Error)?;
+        let mut bfes = vec![];
+        for val in parsed_u64s {
+            if !BFieldElement::is_canonical(val) {
+                return Err(TryFromDigestError::NotCanonical(val));
+            }
+            bfes.push(BFieldElement::from(val))
+        }
```
jan-ferdinand, line 113: I'm a fan of iterator transforms. :blush: This is an opinion and not necessarily a change request.

```suggestion
        let bfe_try_from = |v: u64| -> Result<BFieldElement, _> {
            let bfe = BFieldElement::is_canonical(v).then(|| BFieldElement::new(v));
            bfe.ok_or(TryFromDigestError::NotCanonical(v))
        };
        let bfes: Vec<_> = parsed_u64s.into_iter().map(bfe_try_from).try_collect()?;
```

Upsides: :sunglasses:.
Downsides: not as early a return (I think).

twenty-first/src/math/digest.rs
```diff
@@ -145,22 +153,26 @@ impl From<Digest> for [u8; Digest::BYTES] {
...
+        fn chunk_into_bfe(chunk: &[u8]) -> Result<BFieldElement, TryFromDigestError> {
             let mut arr = [0u8; BFieldElement::BYTES];
             arr.copy_from_slice(chunk);
-            BFieldElement::from(arr)
-        };
+            let int = u64::from_le_bytes(arr);
+            match BFieldElement::is_canonical(int) {
+                true => Ok(BFieldElement::from(arr)),
+                false => Err(TryFromDigestError::NotCanonical(int)),
+            }
```
jan-ferdinand, line 167: I think I prefer this over matching on a `bool`[^1].

```suggestion
            BFieldElement::is_canonical(int)
                .then(|| BFieldElement::new(int))
                .ok_or(TryFromDigestError::NotCanonical(int))
```

[^1]: I have used such matches in the past but dislike their readability now.

jan-ferdinand (approved):

twenty-first/src/error.rs
```diff
@@ -39,13 +39,16 @@ pub enum TryFromDigestError {
     #[error("invalid `BFieldElement`")]
     InvalidBFieldElement(#[from] ParseBFieldElementError),
 
+    #[error("Not Canonical. ({0})")]
```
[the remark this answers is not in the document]
dan-da: sure, works for me.

twenty-first/src/math/digest.rs
dan-da: that's a clever transform.  I didn't immediately see an interator solution, hence the mutating for loop.

The tradeoff I see is that your solution is probably more rust-idiomatic and doesn't require a `mut`, while the for loop is more readable for people like myself coming from procedural languages.   I really have to study your transform to understand what it is doing, but it's good exercise for me, so I thank-you for making it.

I will use your version and add a comment to explain it's purpose.

early-return or not doesn't seem a concern with only 5 iterations.

twenty-first/src/math/digest.rs
dan-da: ok, will do.

PR #204, opened 2024-04-25
fix!: Add field-length indicator to BFE encoding of polynomial

### LINKED ISSUE #203: Manual `BfieldCodec` implementation for `Polynomial` is not consistent with a derived one
In #200 we manually made a `BFieldCodec` implementation for `Polynomial`. But this implementation is inconsistent with a derived one that would have a size indicator for the coefficients field. This creates problems downstream, and especially the [`tasm-lang`](https://github.com/TritonVM/tasm-lang) would have to special-case a lot on `Polynomial` and basically make it into an atomic/native type for this to work.
```rust
pub struct Polynomial<FF: FiniteField> {
    pub coefficients: Vec<FF>,
}
```

The solution is to keep the manual implementation (because we need to remove trailing zeros from the encoding to easily read the degree of an encoded polynomial in Triton VM) but change it such that it carries the length of the field.

Sword-Smith:
This field-length value is needed in the encoding to allow downstream dependencies to treat an encoded polynomial stored in Triton VM RAM as any other struct, since all encoded struct are assumed to be encoded consistently with a derived `BFieldCodec` implementation.

This closes #203.

### DIFF
--- a/twenty-first/src/math/bfield_codec.rs
+++ b/twenty-first/src/math/bfield_codec.rs
@@ -1,3 +1,4 @@
+use std::cmp::Ordering;
 use std::error::Error;
 use std::fmt::Debug;
 use std::fmt::Display;
@@ -391,16 +392,13 @@ impl<T: BFieldCodec + FiniteField> BFieldCodec for Polynomial<T> {
             return Err(Self::Error::Other(BFieldCodecError::InvalidLengthIndicator));
         };
 
+        // Indicated sequence length is 1 + field size, as the size indicator takes up 1 word.
         let indicated_sequence_length = coefficients_field_length_indicator + 1;
-        if sequence.len() < indicated_sequence_length {
-            return Err(Self::Error::Other(BFieldCodecError::SequenceTooShort));
-        }
-
-        if sequence.len() > indicated_sequence_length {
-            return Err(Self::Error::Other(BFieldCodecError::SequenceTooLong));
-        }
-
-        let decoded_vec = Vec::<T>::decode(&sequence[1..])?;
+        let decoded_vec = match sequence.len().cmp(&indicated_sequence_length) {
+            Ordering::Equal => Vec::<T>::decode(&sequence[1..]),
+            Ordering::Less => Err(BFieldCodecError::SequenceTooShort),
+            Ordering::Greater => Err(BFieldCodecError::SequenceTooLong),
+        }?;
 
         let encoding_contains_trailing_zeros = decoded_vec
             .last()
@@ -543,6 +541,7 @@ mod tests {
     use proptest::collection::size_range;
     use proptest::collection::vec;
     use proptest::prelude::*;
+    use proptest::test_runner::TestCaseResult;
     use proptest_arbitrary_interop::arb;
     use test_strategy::proptest;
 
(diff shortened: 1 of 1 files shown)


twenty-first/src/math/bfield_codec.rs
```diff
@@ -373,14 +375,32 @@ pub enum PolynomialBFieldCodecError {
...
+        if sequence.is_empty() {
+            return Err(Self::Error::Other(BFieldCodecError::EmptySequence));
+        }
+
+        let coefficients_field_length_indicator: Result<usize, _> = sequence[0].value().try_into();
+        let Ok(coefficients_field_length_indicator) = coefficients_field_length_indicator else {
+            return Err(Self::Error::Other(BFieldCodecError::InvalidLengthIndicator));
+        };
+
+        let indicated_sequence_length = coefficients_field_length_indicator + 1;
```
jan-ferdinand, line 394: I'm not sure what the `+ 1` is for. Add a comment?

twenty-first/src/math/bfield_codec.rs
```diff
@@ -373,14 +375,32 @@ pub enum PolynomialBFieldCodecError {
...
+        if sequence.len() < indicated_sequence_length {
+            return Err(Self::Error::Other(BFieldCodecError::SequenceTooShort));
+        }
+
+        if sequence.len() > indicated_sequence_length {
+            return Err(Self::Error::Other(BFieldCodecError::SequenceTooLong));
+        }
+
+        let decoded_vec = Vec::<T>::decode(&sequence[1..])?;
```
jan-ferdinand, line 403: Possibly:
```suggestion
        let decoded_vec = match sequence.len().cmp(&indicated_sequence_length) {
            Ordering::Equal => Vec::<T>::decode(&sequence[1..]),
            Ordering::Less => Err(BFieldCodecError::SequenceTooShort),
            Ordering::Greater => Err(BFieldCodecError::SequenceTooLong),
        }?;
```

twenty-first/src/math/bfield_codec.rs
```diff
@@ -732,58 +762,60 @@ mod tests {
...
-    #[proptest]
-    fn disallow_trailing_zeros_in_poly_encoding_bfe(
-        mut polynomial: Polynomial<BFieldElement>,
-        #[filter(!#leading_coefficient.is_zero())] leading_coefficient: BFieldElement,
-        #[strategy(1usize..30)] num_leading_zeros: usize,
+    fn disallow_trailing_zeros_in_poly_encoding_prop<T: FiniteField + BFieldCodec>(
+        mut polynomial: Polynomial<T>,
+        leading_coefficient: T,
+        num_leading_zeros: usize,
     ) {
```
jan-ferdinand, line 769: I like this. If you make the return type a `PropTestResult`, you can even keep using `prop_assert_eq!` and `prop_assert!`.
```suggestion
    fn disallow_trailing_zeros_in_poly_encoding_prop<T: FiniteField + BFieldCodec>(
        mut polynomial: Polynomial<T>,
        leading_coefficient: T,
        num_leading_zeros: usize,
    ) -> TestCaseResult {
```

It would also require
- returning `Ok(())` in this function, and
- changing the two `proptest`s that use this function to conclude with `?;`.

jan-ferdinand (approved):
Good stuff. I left some suggestions in-line.

twenty-first/src/math/bfield_codec.rs
Sword-Smith: Why not?

twenty-first/src/math/bfield_codec.rs
Sword-Smith: Done

Issue #205, opened 2024-04-26
Inconsistent byte encoding for `Digest` on big-endian host-machines


Sword-Smith:
Conversion from `Digest` to `[u8, 40]` happens through:

```rust
impl From<Digest> for [u8; Digest::BYTES] {
    fn from(item: Digest) -> Self {
        let u64s = item.0.iter().map(|x| x.value());
        u64s.map(|x| x.to_ne_bytes())
            .collect::<Vec<_>>()
            .concat()
            .try_into()
            .unwrap()
    }
}
```

To get the same byte encoding on all platforms, we should use `x.to_le_bytes()` instead, or at least be consistent with the opposite conversion. The opposite conversion currently uses `u64::from_le_bytes(arr)`, so we might as well stay with little-endian encoding unless someone has a pronounced opinion about this. 

The fix must include a test the verifies that digest conversion to and back from an array of 40 bytes is the identity operation. This new test should be added as a `proptest` using the `#[proptest]` pragma from the `proptest` crate that's already a dependency.

PR #206, opened 2024-05-05
revisit `BFieldElement`'s to / from native endian bytes

jan-ferdinand:
This PR fixes the following issues with current conversion from “native” endian bytes to base field elements:

- Make the function infallible. Before, if the passed-in slice was of the incorrect length, the function would panic – see [`.copy_from_slice()`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice). This behavior was not documented.
- Add a reciprocal method to get a “native” endianess byte array from a base field element.
- Test the behavior.

All that said, I think there are potential footguns regarding cross-platform support lingering here. I would argue we should drop “native” endianness in favor of either little or big endian representation, or maybe drop all of that since we already have `from_raw_bytes()` and `.raw_bytes()`.

### DIFF
--- a/twenty-first/src/math/b_field_element.rs
+++ b/twenty-first/src/math/b_field_element.rs
@@ -292,11 +292,13 @@ impl BFieldElement {
         acc
     }
 
-    /// Convert a `BFieldElement` from a byte slice in native endianness.
-    pub fn from_ne_bytes(bytes: &[u8]) -> BFieldElement {
-        let mut bytes_copied: [u8; 8] = [0; 8];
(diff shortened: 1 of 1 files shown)


Sword-Smith:
I totally agree that we shouldn't even have the option of encoding from/to native endianness. I also remarked that in an issue, as it seems we're actually using native endianness at least one place in our code base: https://github.com/Neptune-Crypto/twenty-first/issues/205

jan-ferdinand:
As of #235, `BFieldElement::from_ne_bytes()` is deprecated and will be deleted in one of the next versions.

PR #215, opened 2024-05-24
Fix Fast Reduce

aszepieniec:
This PR fixes `fast_reduce`.

Function `fast_reduce` did not work when the modulus had structure, in particular when it had trailing zeros. The issue was that `structured_multiple` lost track of trailing zeros in the course of reversing coefficient vectors.


jan-ferdinand:
Thanks for identifying and fixing the underlying issue. :clap: I think that would have taken me a very long time (and hope this was not the case for you).

PR #218, opened 2024-06-21
refactor: Make `Digest::LEN` an associated const

jan-ferdinand:
Also, deprecate `digest::DIGEST_LENGTH`.

I think this design is cleaner. It's a bit of a bikeshed. :shrug: Open to suggestions.


aszepieniec (changes requested):
Much cleaner, yes and thanks ... but why do the tests fail?

jan-ferdinand:
> why do the tests fail?

Because I messed up a doc link. Fixed now. :+1:

aszepieniec (approved):

jan-ferdinand:
If I understand correctly, some randomized tests hit large swaths of code only occasionaly, leading to a rather big variance in code coverage. This feels like a separate issue.

PR #219, opened 2024-06-25
Thv/remove leaf idx from mmr mp

Sword-Smith:
- Remove leaf-index from MMR's `MmrMembershipProof`
- Add struct `LeafMutation` for encapsulating the data required to mutate a leaf in an MMR-accumulator
- Change return type of archival-version of MMR (here called `MockMmr`) to only return an MMR membership proof, not the membership proof, *and* the peaks
- change MMR trait method name from `get_peaks` to `peaks`
- change MMR trait method name from `count_leaves` to `num_leaves`.

Not addressed:
Simplification of tests. The tests are very thorough but there is a lot of repetition in them. The number of tests can probably be halved without affecting test coverage negatively.

### DIFF
--- a/twenty-first/src/util_types/mmr/mmr_membership_proof.rs
+++ b/twenty-first/src/util_types/mmr/mmr_membership_proof.rs
@@ -339,7 +339,10 @@ impl<H: AlgebraicHasher> MmrMembershipProof<H> {
         modified
     }
 
-    /// Update a membership proof with a `leaf_mutation` proof.
+    /// Update a membership proof with a `leaf_mutation` data structure. In
+    /// other words: Given a valid MMR membership proof for an MMR, calculate
+    /// the updated MMR membership proof after one of the MMR's leafs have been
+    /// changed.
     pub fn update_from_leaf_mutation(
         &mut self,
         own_mp_leaf_index: u64,
--- a/twenty-first/src/util_types/mmr/mmr_membership_proof.rs
+++ b/twenty-first/src/util_types/mmr/mmr_membership_proof.rs
@@ -771,7 +771,7 @@ mod mmr_membership_proof_test {
         type H = Tip5;
         let leafs: Vec<Digest> = random_elements(leaf_count);
         let archival_mmr: MockMmr<H> = get_mock_ammr_from_digests(leafs.clone());
-        let peaks = archival_mmr.get_peaks();
+        let peaks = archival_mmr.peaks();
         for (i, leaf) in leafs.into_iter().enumerate() {
             let leaf_index = i as u64;
             let mp = archival_mmr.prove_membership(leaf_index);
@@ -820,8 +820,8 @@ mod mmr_membership_proof_test {
             mp.verify(
                 mp_leaf_index as u64,
                 leaf_hash,
-                &archival_mmr.get_peaks(),
-                archival_mmr.count_leaves(),
+                &archival_mmr.peaks(),
+                archival_mmr.num_leaves(),
             );
         }
     }
@@ -871,7 +871,7 @@ mod mmr_membership_proof_test {
             assert!(mp.verify(
                 leaf_index,
                 final_leaf_hash,
-                &archival_mmr_final.get_peaks(),
+                &archival_mmr_final.peaks(),
                 total_leaf_count as u64
             ));
         }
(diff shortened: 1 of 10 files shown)


twenty-first/src/util_types/mmr/mmr_membership_proof.rs
```diff
@@ -221,12 +229,32 @@ impl<H: AlgebraicHasher> MmrMembershipProof<H> {
     /// Batch update multiple membership proofs.
     /// Returns the indices of the membership proofs that were modified where index refers
     /// to the order in which the membership proofs were given to this function.
+    /// Panics if `membership_proofs` and `membership_proof_leaf_indices` do not have
+    /// the same length, or if a leaf index is out-of-bounds for the MMR.
     pub fn batch_update_from_append(
         membership_proofs: &mut [&mut Self],
+        membership_proof_leaf_indices: &[u64],
```
aszepieniec, line 236: Why not just `leaf_indices`?

twenty-first/src/util_types/mmr/mmr_membership_proof.rs
```diff
@@ -306,22 +339,19 @@ impl<H: AlgebraicHasher> MmrMembershipProof<H> {
         modified
     }
 
-    /// Update a membership proof with a `leaf_mutation` proof. For the `membership_proof`
-    /// parameter, it doesn't matter if you use the old or new membership proof associated
-    /// with the leaf update, as they are the same before and after the leaf mutation.
+    /// Update a membership proof with a `leaf_mutation` proof.
```
aszepieniec, line 342: Describing a `LeafMutation` as a proof is misleading. At best, it contains a proof -- and the new leaf and index that is also part of this structure, are certainly necessary for the current update.

twenty-first/src/util_types/mmr/mmr_trait.rs
```diff
@@ -1,7 +1,46 @@
+use super::mmr_accumulator::MmrAccumulator;
+use super::mmr_membership_proof::MmrMembershipProof;
 use crate::math::digest::Digest;
 use crate::util_types::algebraic_hasher::AlgebraicHasher;
 
-use super::{mmr_accumulator::MmrAccumulator, mmr_membership_proof::MmrMembershipProof};
+/// A definition of a mutation of a leaf in an MMR
```
aszepieniec, line 6: This documentation comment is non-descriptive. Try explaining, without listing the individual fields or their types, which kind of data this struct encapsulates, and what its instances are for. Avoid the word "mutation" because this word conjures different meanings in the heads of different people. It might also be worth drawing attention to pitfalls or counter-intuitive features -- in this case: the old leaf is not part of this data structure, and its knowledge is not required to apply the mutation.

aszepieniec (approved):
Thanks! This index has been a pebble in my shoe for a while now. Much cleaner now.

I left a couple of remarks inline, but feel free to skip over them if you disagree.

I wonder if now is the right time to remove the generic type parameter `<H : Algebraic Hasher>` since we know we are not going to instantiate this with anything other than Tip5.

About the tests -- I agree that there is a lot of duplication and the test suite could benefit from some cleanup. In fact, a cleanup might result from simply adding proptests -- which is something we could make an issue about and flag as beginner-friendly, as it is an instructive exercise to people who are unfamiliar with the proptest framework.

Lastly, without implying anything about the priorities of oughts, let me use the opportunity to cite for the record two lines from Alan's written style guide:

 - Mutation denotes accruing change passively. Modification implies design and intent.
 - Trees in the forest, and plants in general, have leaves. Merkle trees have leafs.

twenty-first/src/util_types/mmr/mmr_membership_proof.rs
Sword-Smith: I like the descriptive name here. I prefer specifying that the leaf-indices relate to the membership proofs and not, e.g., to the indices of the leafs that are appended.

Sword-Smith:
> Lastly, without implying anything about the priorities of oughts, let me use the opportunity to cite for the record two lines from Alan's written style guide:
>     * Trees in the forest, and plants in general, have leaves. Merkle trees have leafs.

I'm OK changing all occurrences of "leaves" to "leafs".

twenty-first/src/util_types/mmr/mmr_membership_proof.rs
Sword-Smith: Agreed.

twenty-first/src/util_types/mmr/mmr_trait.rs
Sword-Smith: I came up with:
>  A wrapper for the data needed to change the value of a leaf in an MMR when
 only the MMR-accumulator is known, i.e., only the peaks and the leaf-count
are known.

PR #226, opened 2024-08-01
refactor(mmr_successor_proof): Flatten digest list

Sword-Smith:
suggested change to `MmrSuccessorProof`.

Use Merkle tree node index instead of a length of a vector for inner-loop's stop condition. Also make `verify` more structured by reducing use of early return.

### DIFF
--- a/twenty-first/src/util_types/mmr/mmr_successor_proof.rs
+++ b/twenty-first/src/util_types/mmr/mmr_successor_proof.rs
@@ -117,10 +117,14 @@ impl MmrSuccessorProof {
             return false;
         }
 
+        let num_new_peaks: u32 = new_mmra.peaks().len().try_into().unwrap();
+        if new_mmra.num_leafs().count_ones() != num_new_peaks {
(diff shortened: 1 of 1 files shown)


twenty-first/src/util_types/mmr/mmr_successor_proof.rs
```diff
@@ -270,28 +272,21 @@ mod test {
...
-        // Missing path
+        // Missing path element
         if !mmr_successor_proof.paths.is_empty() {
             let mut fake_mmr_successor_proof_4 = mmr_successor_proof.clone();
             fake_mmr_successor_proof_4.paths.pop();
             prop_assert!(!fake_mmr_successor_proof_4.verify(&old_mmr, &new_mmr));
         }
 
-        // One path too many
+        // One element
```
aszepieniec, line 287:
```suggestion
        // One element too many
```

aszepieniec (approved):
LGTM

PR #228, opened 2024-08-05
Working code, small tests

Sword-Smith:
This is how the `RootFromAuthStruct` implementation could look in Rust. Code works for these two very small test cases.

### DIFF
--- a/twenty-first/src/util_types/mmr.rs
+++ b/twenty-first/src/util_types/mmr.rs
@@ -1,4 +1,5 @@
 pub mod mmr_accumulator;
+pub mod mmr_authentication_struct;
 pub mod mmr_membership_proof;
 pub mod mmr_successor_proof;
 pub mod mmr_trait;
--- a/twenty-first/src/util_types/mmr/mmr_authentication_struct.rs
+++ b/twenty-first/src/util_types/mmr/mmr_authentication_struct.rs
@@ -434,6 +434,27 @@ mod tests {
         assert!(authenticated_auth_structs.is_empty());
     }
 
+    #[should_panic]
+    #[test]
+    fn panics_on_missing_nd_digests() {
+        let tree_height = 3u32;
+        let num_leafs = 1u64 << tree_height;
+        let leafs: Vec<Digest> = random_elements(num_leafs.try_into().unwrap());
+        let tree = MerkleTree::<Tip5>::new::<CpuParallel>(&leafs).unwrap();
+
+        let authenticated_auth_struct =
+            AuthStructIntegrityProof::new_from_merkle_tree(&tree, vec![0]);
+        let AuthenticatedMerkleAuthStruct {
+            auth_struct,
+            indexed_leafs,
+            mut witness,
+        } = authenticated_auth_struct;
+        witness.nd_sibling_indices.clear();
+        witness.nd_siblings.clear();
+
+        witness.root_from_authentication_struct(tree_height, auth_struct, indexed_leafs);
+    }
+
     #[test]
     fn auth_struct_from_mmr_mps_test_height_5_9_indices() {
         let local_tree_height = 5;
--- a/twenty-first/src/util_types/mmr/mmr_authentication_struct.rs
+++ b/twenty-first/src/util_types/mmr/mmr_authentication_struct.rs
@@ -26,7 +26,7 @@ const ROOT_MT_INDEX: u64 = 1;
 /// A witness to facilitate the proving of the authenticity of a Merkle
 /// authentication struct.
 #[derive(Debug, Clone)]
-pub struct MerkleAuthenticationStructAuthenticityWitness {
+pub struct AuthStructIntegrityProof {
     // All indices are Merkle tree node indices
     pub nd_auth_struct_indices: Vec<u64>,
     pub nd_sibling_indices: Vec<(u64, u64)>,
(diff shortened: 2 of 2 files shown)


twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -1033,4 +1042,166 @@ pub mod merkle_tree_test {
...
+
+            leaf_xfe_lo + leaf_xfe_hi
+        }
+
+        assert_eq!(leafs.len(), leafs_node_indices.len());
+        assert_eq!(auth_struct.len(), auth_struct_indices.len());
+        let (alpha, gamma, delta) = {
+            let mut sponge = Tip5::init();
+            sponge.pad_and_absorb_all(&leafs.encode());
+            sponge.pad_and_absorb_all(&auth_struct.encode());
```
aszepieniec, line 1067: This is a non-standard mode of use of the sponge construction. While I cannot find an attack, I am concerned that a vulnerability might hide here. If we definitely want this construction (for instance, because it is more efficient to compute in Triton VM) then we need an iron-clad security argument.

The alternative would be something like `sponge.pad_and_absorb_all(&(leafs, auth_struct).encode())`.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -1033,4 +1042,166 @@ pub mod merkle_tree_test {
...
+            let leaf_index_xfe =
+                XFieldElement::new([leaf_index_hi, leaf_index_lo, BFieldElement::zero()]);
+            let fact = leafs_as_xfes[i] - gamma + delta * leaf_index_xfe;
+            p *= fact;
+        }
+        for i in 0..auth_struct.len() {
+            let leaf_index_hi = bfe!(auth_struct_indices[i] >> 32);
+            let leaf_index_lo = bfe!(auth_struct_indices[i] & u32::MAX as u64);
+            let leaf_index_xfe =
+                XFieldElement::new([leaf_index_hi, leaf_index_lo, BFieldElement::zero()]);
```
aszepieniec, line 1091: Why are you putting the more significant limb (`hi`) on the least significant coefficient, but nothing on the most significant coefficient? It's counter-intuitive.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -1033,4 +1042,166 @@ pub mod merkle_tree_test {
...
+            parent_index_xfe = left_index_xfe / xfe!(2);
+
+            let fact1 = l_xfe - gamma + delta * left_index_xfe;
+            let fact2 = r_xfe - gamma + delta * right_index_xfe;
+            let fact_parent = t_xfe - gamma + delta * parent_index_xfe;
+
+            p *= fact1.inverse() * fact2.inverse() * fact_parent;
+        }
+
+        let x_as_xfe = XFieldElement::new([
```
aszepieniec, line 1125: A better name might be `index_1_as_xfe`.

aszepieniec: Or `root_index_as_xfe`.

aszepieniec (commented):
The algorithm looks correct to me. I left a couple of nits inline.

Do you want to merge this into `twenty-first`? The repository doesn't have any other functions that take nondeterminism. You can make it fit though by planting the relevant function on a data structure, similarly to what was done with `MmrSuccessorProof`. In this case the right struct might be `AuthStructIntegrityProof` or something like that, and the function `verify(&self, leafs : &[Digest], leaf_indices: &[u64], auth_struct: &[Digest], root: Digest) -> bool`.

I intentionally left the indices of the auth path elements from this function signature because they can be inferred from the leaf indices. They are not part of the current authentication structure. Come to think of it, rather than computing the auth struct indices, it probably makes more sense to divine them in and then verify them.

Since the indices are `u64`s, this method and/or struct probably belongs in `mmr/` rather than `merkle_tree.rs`.

twenty-first/src/util_types/merkle_tree.rs
Sword-Smith: Your suggestion probably requires allocation in Triton VM. It's preferable if we can avoid that as much as possible.

Sword-Smith:
Moved code to `mmr` per your suggestion. auth-struct indices are now divined in, not part of the public data. Also added many tests.

Sword-Smith:
So we have a nice solution that works for Merkle trees. How do we translate this to something that works for Merkle mountain ranges? Do we just include an authentication structure for each peak that the list of leafs are contained in?

aszepieniec:
> So we have a nice solution that works for Merkle trees. How do we translate this to something that works for Merkle mountain ranges? Do we just include an authentication structure for each peak that the list of leafs are contained in?

I don't see how things could be done more efficiently, seen as how the various trees in an MMR are completely uncorrelated.

Sword-Smith:
Now this data structure and associated data can be derived from two sources: Either from an (MMR, membership-proof list) pair, or from a Merkle tree and a list of indices into this tree.

Sword-Smith:
I'm happy with the state of this code now. Now the only question is just whether a) we actually want to use this data structure, and b) if we do want to use it, whether it belongs in this repo.

Oh, and c) whether the interface of `root_from_authentication_struct` is conducive to a short execution trace in Triton VM.

jan-ferdinand:
@Sword-Smith, is this PR still relevant?

PR #231, opened 2024-09-28
switch to get-size2

bircni:


aszepieniec:
(Why) would you say that getsize2 is superior to getsize?

bircni:
https://github.com/DKerp/get-size/issues/22

Sword-Smith:
@bircni Thanks for the PR! This seems like a good upgrade, that we should include in the 0.43 version and downstream.

bircni:
would be nice!

bircni:
fixed the merge conflict

Sword-Smith:
Thanks. Merged.

PR #233, opened 2024-09-30
chore: Bump all dependencies

Sword-Smith:
Bumps all dependencies to latest version, or newer.

### DIFF
--- a/bfieldcodec_derive/Cargo.toml
+++ b/bfieldcodec_derive/Cargo.toml
@@ -6,9 +6,9 @@ license = "GPL-2.0"
 description = "Derive macro for BFieldCodec."
 
 [dependencies]
-syn = { version = "^2", features = ["full", "derive"] }
-quote = "1.0"
-proc-macro2 = "1.0"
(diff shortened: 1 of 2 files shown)


jan-ferdinand (approved):
I'm not opposing this change, since “just use the latest” has been our approach to dependencies for a long time now. However, given that you're asking for my review right now, the recommendation given in “Rust for Rustaceans”[^rfr] is:

> The right strategy is to list the _earliest_ version that has all the things your crate depends on […].

The subsection also contains the explanation why this is the right strategy, how to accomplish it, and how to enforce it in CI.[^lend]

[^rfr]: Chapter 5, Section “Versioning”, Subsection “Minimal Dependency Versions”, emphasis mine
[^lend]: I don't want to paste the entire section on the public internet but am happy to lend you the book should you be interested.

Sword-Smith:
Thanks for the insight. As @jan-ferdinand explained, the reason not to bump versions if stuff works is that it'll be harder for other crates to share dependencies with this.

PR #234, opened 2024-10-02
Support more types in `BFieldCodec` and in `BFieldElement` conversion

jan-ferdinand:
Implement `BFieldCodec` for `u8` and `u16`, and many type conversions to and from `BFieldElement`.

### DIFF
--- a/twenty-first/src/math/b_field_element.rs
+++ b/twenty-first/src/math/b_field_element.rs
@@ -397,39 +397,6 @@ impl FromStr for BFieldElement {
     }
 }
 
-impl From<u8> for BFieldElement {
-    fn from(value: u8) -> Self {
-        Self::new(value.into())
-    }
-}
-
-impl From<u16> for BFieldElement {
-    fn from(value: u16) -> Self {
-        Self::new(value.into())
-    }
-}
-
-impl From<u32> for BFieldElement {
-    fn from(value: u32) -> Self {
-        Self::new(value.into())
-    }
-}
-
-impl From<u64> for BFieldElement {
-    fn from(value: u64) -> Self {
-        Self::new(value)
-    }
-}
-
-impl From<i32> for BFieldElement {
-    fn from(value: i32) -> Self {
-        match value {
-            v if v >= 0 => Self::new(v as u64),
-            v => -Self::new(-v as u64),
-        }
-    }
-}
-
 impl From<usize> for BFieldElement {
     fn from(value: usize) -> Self {
         // targets with wider target pointers don't exist at the time of writing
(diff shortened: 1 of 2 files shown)


Sword-Smith (changes requested):
Apart from my comments on the code, let's also add some tests. I came up with these:

```rust
    #[proptest]
    fn conversion_from_u128_is_correct(v: u128) {
        let bfe = BFieldElement::from(v);
        prop_assert_eq!(v % (BFieldElement::P as u128), bfe.value() as u128);
    }

    #[proptest]
    fn conversion_from_i128_is_correct(v: i128) {
        let v = if v < 0 {
            (v as u128) - BFieldElement::R2 as u128
        } else {
            v as u128
        };
        let bfe = BFieldElement::from(v);
        prop_assert_eq!(v % (BFieldElement::P as u128), bfe.value() as u128);
    }

    #[test]
    fn can_convert_from_minus1_i128() {
        let bfe = BFieldElement::from(-1i128);
        assert_eq!(BFieldElement::MAX, bfe.value());
    }
```

jan-ferdinand:
Thanks for the feedback! The speed improvements are definitely worth including.

Regarding the suggested tests:
- The third of the suggested tests is already included, as one of the lines in function `bfield_element_can_be_converted_to_and_from_many_types`.
- The first two tests feel like a repetition of the suggested conversion code, and so I'm not sure they are meaningful. How about the following two tests instead?

```rust
#[proptest]
fn naive_and_actual_conversion_from_u128_agree(v: u128) {
    fn naive_conversion(x: u128) -> BFieldElement {
        let p = BFieldElement::P as u128;
        let value = (x % p) as u64;
        BFieldElement::new(value)
    }

    prop_assert_eq!(naive_conversion(v), BFieldElement::from(v));
}

#[proptest]
fn naive_and_actual_conversion_from_i128_agree(v: i128) {
    fn naive_conversion(x: i128) -> BFieldElement {
        let p = BFieldElement::P as i128;
        let value = x.rem_euclid(p) as u64;
        BFieldElement::new(value)
    }

    prop_assert_eq!(naive_conversion(v), BFieldElement::from(v));
}
```

Sword-Smith:
I like those new test suggestions.

jan-ferdinand:
The previous implementations `From<$t>` for signed types `$t` suffered from incompleteness in the case of `$t::MIN`. With the latest commit, all signed types delegate to `i128`, which in turns has one branch and then delegates to `u128`. Let me know what you think of that. :relaxed:

Sword-Smith (approved):

PR #235, opened 2024-10-03
Move canon check from `Digest` to `BFieldElement`

jan-ferdinand:
- Reject non-canonical `BFieldElement`s when parsing string slices or interpreting byte sequences
- Move canon check of `BFieldElement`s from `Digest` to `BFieldElement`
- Deprecate functionality for platform-dependent endianness when decoding

BREAKING CHANGE: The error variant `NotCanonical` from enum `TryFromDigestError` has moved into enum `ParseBFieldElementError`.

This was described as the cleaner solution in #199, a sentiment I concur with.

Supersedes #206.

### DIFF
--- a/twenty-first/src/util_types/mmr/mmr_accumulator.rs
+++ b/twenty-first/src/util_types/mmr/mmr_accumulator.rs
@@ -459,15 +459,10 @@ pub mod util {
 
 impl<'a> Arbitrary<'a> for MmrAccumulator {
     fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
-        let mut buffer = [0u8; 8];
-        u.fill_buffer(&mut buffer)?;
-        let num_leafs = u64::from_be_bytes(buffer) >> 1; // num_leafs can be at most 63 bits
-
-        let mut peaks = vec![];
-        for _ in 0..(num_leafs.count_ones()) {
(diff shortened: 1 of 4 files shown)


twenty-first/src/util_types/mmr/mmr_accumulator.rs
```diff
@@ -459,15 +459,10 @@ pub mod util {
...
-
-        let mut peaks = vec![];
-        for _ in 0..(num_leafs.count_ones()) {
-            let peak = Digest::arbitrary(u)?;
-            peaks.push(peak);
-        }
+        let num_leafs = u.arbitrary::<u64>()? >> 1; // num_leafs can be at most 63 bits
+        let peaks = (0..num_leafs.count_ones())
+            .map(|_| Digest::arbitrary(u))
+            .try_collect()?;
```
jan-ferdinand, line 465: This piece of code came up when searching for the regex `(b|l|n)e_bytes`. The change doesn't really belong in this PR, but I couldn't resist simplifying things.

jan-ferdinand, line 465: This piece of code came up when searching for the regex `(b|l|n)e_bytes`. The change doesn't really belong in this PR, but I couldn't resist simplifying things.

jan-ferdinand:
@Sword-Smith, any opposition?

PR #237, opened 2024-10-18
Optionally borrow polynomial coefficients

jan-ferdinand:
Refactor polynomials to (a) make their innards private and (b) turn their innards into a clone-on-write slice, reducing the number of allocations needed for read-only operations like polynomial evaluation.

### Pros
- Certain functions, like evaluation, can happen without owning the coefficients, which reduces memory consumption.
- There are fewer sneaky memory allocations, _i.e._, it is more obvious in calling code whether allocations are happening. [^exceptions]

### Cons
- Explicit lifetimes are required more often.
- Polynomial's implementations of traits `Zero` and `One` are less usable now. Should `poly.is_zero()` be unusable due to lifetime constraints (“`poly` escapes the method body”), you can use `poly.degree() < 0` instead.

A lot of lines have changed only because of re-grouping to avoid multiple `impl` blocks with the same trait & lifetime bounds. Commit 65914d5a736cce472080b12a168b4ef5d22820f only makes the polynomial's innards private to simplify review.

## Performance

The benchmark suite indicates some improvements, some regressions, and I can't seem to identify clear patterns. I will try to make sense of the results and report back in a separate comment.

[^exceptions]: Methods taking `&mut self`, like `scalar_mul_mut()` or `normalize()`, are unfortunate exceptions if called on a `Polynomial::new_borrowed()`. I thought about making those methods consuming, and return a `Polynomial<'static, FF>`, but have decided against that for now. Let me know if you disagree with that decision.

### DIFF
--- a/twenty-first/benches/formal_power_series_inverse.rs
+++ b/twenty-first/benches/formal_power_series_inverse.rs
@@ -23,7 +23,11 @@ fn fpsi(c: &mut Criterion) {
 
         let id = BenchmarkId::new("fpsi", format!("2^{log2_degree}"));
         group.bench_function(id, |b| {
-            b.iter(|| polynomial.formal_power_series_inverse_newton(1 << (log2_degree + 1)))
+            b.iter(|| {
+                let precision = 1 << (log2_degree + 1);
+                let polynomial = polynomial.clone();
+                polynomial.formal_power_series_inverse_newton(precision)
+            })
         });
     }
     group.finish();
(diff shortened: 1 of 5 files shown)


jan-ferdinand:
@aszepieniec, @Sword-Smith, I have requested both of you for review only because I'm not sure who this is better suited for. :relieved:

Sword-Smith:
Let me know if/when you need me to run benchmarks.

aszepieniec (approved):
:cow:

jan-ferdinand:
# Performance

- scaling a base-field polynomial by a base-field scalar is faster by about 9% (in use)
- constructing a zerofier the “smart” way is slower by about 2% (not in use)
- constructing a zerofier the “fast” way is faster for some instances, by about 5% (in use)
- constructing a zerofier the “naive” way is faster for larger instances, by up to 8% for the largest instance (not in use)
- other performance changes are less than 2%
- most of those other performance changes are not consistent across instance sizes, making them less relevant in my eyes

In my eyes, this PR is ready to be merged.

PR #240, opened 2024-12-18
docs: Describe Merkle Mountain Ranges

jan-ferdinand:
Includes examples and the terminology for the various kinds of indices.

@Sword-Smith, did I remember everything correctly?


twenty-first/src/util_types/mmr/mmr_trait.rs
```diff
@@ -39,42 +39,129 @@ impl LeafMutation {
...
+/// A “Merkle Mountain Range” (MMR) is a data structure for storing lists of
+/// hashes. Similar to [Merkle trees](crate::prelude::MerkleTree), they enable
+/// efficient set inclusion proofs. Unlike Merkle trees, the number of leafs of
+/// an MMR is not required to be a power of 2.
+///
+/// One valid perspective of this data structure is to see it as a list of
+/// Merkle trees. Each of the Merkle tree's roots is one peak of the MMR.
+/// In order to keep the number of peaks to a minimum, any two neighboring
+/// Merkle trees of the same height are merged, which results in a single tree
+/// that has one additional “tier”.
```
Sword-Smith, line 51: I guess "neighboring" is superfluous here. Any two Merkle trees of the same height are merged.

Sword-Smith: "tier", I guess, can also be described as height.

twenty-first/src/util_types/mmr/mmr_trait.rs
```diff
@@ -39,42 +39,129 @@ impl LeafMutation {
...
+/// ```markdown
+///       _                        _
+///      ╱  ╲                     ╱  ╲
+///     ╱    ╲                   ╱    ╲
+///    _      _     _     _     _      _
+///   ╱ ╲    ╱ ╲   ╱ ╲   ╱ ╲   ╱ ╲    ╱ ╲
+///  _   _  _   _ _   _ _   _ _   _  _   _
+/// ```
+///
+/// It has multiple problems: There are two neighboring Merkle trees of equal
```
Sword-Smith, line 106: It's enough of a violation that there are two Merkle trees of equal height, they don't have to be neighboring.

twenty-first/src/util_types/mmr/mmr_trait.rs
```diff
@@ -39,42 +39,129 @@ impl LeafMutation {
...
+/// usual Merkel tree indexing is used to identify nodes. This is called the
+/// “Merkle tree index” and is usually only meaningful in combination with a
+/// peak index, for example like [here][mt_peak_idx].
+///
+/// [mt_proof]: crate::util_types::merkle_tree::MerkleTree::authentication_structure
+/// [mt_peak_idx]: crate::util_types::mmr::shared_basic::leaf_index_to_mt_index_and_peak_index
 pub trait Mmr {
     // constructors cannot be part of the interface since the archival version requires a
     // database which we want the caller to create, and the accumulator does not need a
     // constructor.
```
Sword-Smith, line 131: This comment is deprecated, as the archival version of the MMR was moved downstream, to `neptune-core`. This was done because all disk operations were made async such that the archival version could not share a trait (interface) with the accumulator-version.

twenty-first/src/util_types/mmr/mmr_trait.rs
```diff
@@ -39,42 +39,129 @@ impl LeafMutation {
...
+///         ──── 14 ───                           ╮
+///        ╱           ╲                          │
+///       6             13            21          │
+///      ╱  ╲          ╱  ╲          ╱  ╲         ├╴ node indices
+///     ╱    ╲        ╱    ╲        ╱    ╲        │
+///    2      5      9      12     17     20      │
+///   ╱ ╲    ╱ ╲    ╱ ╲    ╱ ╲    ╱ ╲    ╱ ╲      │
+///  0   1  3   4  7   8  10 11  15 16  18 19 22  ╯
+///
+///  0   1  2   3  4   5  6   7  8   9  10 11 12  ←── leaf indices
```
Sword-Smith, line 68: The node indices here are off-by-one compared to the convention we're using. The leaf indices are correct.

Sword-Smith (changes requested):
It seems that you have an off-by-one error in the values of all node-indices. We follow the same convention as in this document, where the lowest node index is 1, not 0:
https://www.chainsecurity.com/blog/merkle-mountain-range-mmr-the-case-of-herodotus

From an algorithmical perspective, I think it's always preferable to only consider leaf indices and Merkle-tree indices (into the local Merkle tree) as these mappings can be calculated in constant time. But in some cases, you do need the concept of a node index as well.

I also have one objection, wrt. the use of "neighboring".

PR #243, opened 2024-12-23
fix: typos in documentation files

vtjl10:
Corrected `buildings` to `building`


jan-ferdinand:
Thanks for pointing this out. :+1: I have incorporated a more general change.

PR #244, opened 2024-12-26
perf(MmrSuccessorProof): Use shorter auth paths

jan-ferdinand:
Improve the average & worst-case time & memory performance of verifying that one MMR Accumulator is the successor of some other MMR Accumulator, and document the new algorithm.

An explanation of the verification algorithm follows. As an example, take the following Merkle Mountain Range with 42 = 101010₂ leafs. Peaks are marked with `·`s.

```
                               ·
               ╭───────────────┴───────────────╮
       ╭───────┴───────╮               ╭───────┴───────╮               ·
   ╭───┴───╮       ╭───┴───╮       ╭───┴───╮       ╭───┴───╮       ╭───┴───╮
 ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ·
╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮
```

Adding 8 leafs to it results in the following new Merkle Mountain Range with 50 = 110010₂ leafs. The addition is drawn boxy and bold. The elements of the required authentication path is marked with `x`es. The new peaks are marked with `○`. The one peak that's both old and new is marked `⊙`.

```
            ⊙
  ╭─────────┴─────────╮                               ○
┄┄┴┄┄         ╭───────┴───────╮               ·━━━━━━━┻━━━━━━━┓
          ╭───┴───╮       ╭───┴───╮       ╭───┴───╮       ┏━━━┻━━━x
        ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ╭─┴─╮   ·━┻━x   ┏━┻━┓   ○
       ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ╭┴╮ ┏┻┓ ┏┻┓ ┏┻┓ ┏┻┓
```

The new MMR Accumulator (the one with 50 leafs) is a successor to the old MMR Accumulator (the one with 42 leafs) if all of the following are true:
- Both MMR Accumulators are self-consistent.
- The new MMR Accumulator has more or equal the number of leafs as the old one.
- All “shared peaks” are identical. In the example, only the highest peak is shared.
- The first unshared peak in the new MMR Accumulator is a Merkle tree that contains all remaining old peaks. In the example, the second new peak contains the second and third old peaks.

All subsequent new peaks are irrelevant for verification. In the example, this is only the smallest peak of the new MMR Accumulator.

To re-compute the first unshared peak in the new MMR Accumulator,
- start the Merkle tree authentication from the correct height, that is, from the height of the smallest old peak;
- for any left sibling, use an old peak; and
- for any right sibling, use an element of the authentication path.

### DIFF
--- a/twenty-first/src/error.rs
+++ b/twenty-first/src/error.rs
@@ -9,6 +9,11 @@ use crate::prelude::x_field_element::EXTENSION_DEGREE;
 use crate::prelude::BFieldElement;
 pub use crate::util_types::merkle_tree::MerkleTreeError;
 
+pub(crate) const USIZE_TO_U64_ERR: &str =
+    "internal error: type `usize` should have at most 64 bits";
+pub(crate) const U32_TO_USIZE_ERR: &str =
+    "internal error: type `usize` should have at least 32 bits";
+
 #[derive(Debug, Clone, Eq, PartialEq, Error)]
 #[non_exhaustive]
 pub enum ParseBFieldElementError {
--- a/twenty-first/src/util_types/mmr.rs
+++ b/twenty-first/src/util_types/mmr.rs
@@ -4,3 +4,6 @@ pub mod mmr_successor_proof;
 pub mod mmr_trait;
 pub mod shared_advanced;
 pub mod shared_basic;
+
+const TOO_MANY_LEAFS_ERR: &str =
+    "internal error: Merkle Mountain Ranges should have at most 2^63 leafs";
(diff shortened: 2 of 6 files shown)


aszepieniec (approved):
The algorithm and the code look good to me. Nice find!

PR #247, opened 2025-01-08
feat: Provide sequential Merkle tree builder

jan-ferdinand:
Previously, the only way to construct a Merkle tree was to use `CpuParallel`. However, in some cases, using parallelism does not make sense or even adversely effects the program.

Add the Merkle tree maker `CpuSequential`, which constructs the entire tree sequentially.

Please note that at first glance, it might seem as if this PR increases the `DEFAULT_PARALLELIZATION_CUTOFF`. However, this change is merely to offset the changes around it's only use I introduce in this PR.

I'm not fully convinced that the new struct `CpuSequential` should go into the crate's prelude. Opinions welcome.


Sword-Smith (approved):
Looks good. One comment: Maybe also verify that the roots agree in the tests, where you Merkle trees built in parallel and serial? To me, it's not obvious what equality for two Merkle trees mean. If the tests compared the roots, that doubt would be removed from my mind.

I don't have an opinion on the prelude question.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -689,6 +706,23 @@ pub mod merkle_tree_test {
...
     }
 
+    #[proptest]
+    fn merkle_tree_makers_behave_identically_on_random_input(leafs: Vec<Digest>) {
+        let parallel = MerkleTree::new::<CpuParallel>(&leafs);
+        let sequential = MerkleTree::new::<CpuSequential>(&leafs);
+        prop_assert_eq!(parallel, sequential);
+    }
+
+    #[proptest]
```
Sword-Smith, line 716: cases here, could be limited to 10 or 20, I think.

jan-ferdinand: The tests finishes in less than 2 seconds on my laptop. From this perspective, I don't think the current number of cases is a problem, but maybe there are other considerations to take into account?

Sword-Smith: If it's that fast, don't bother changing anything.

jan-ferdinand:
> To me, it's not obvious what equality for two Merkle trees mean.

`MerkleTree` derives `Eq`, so equality is deferred to the underlying `nodes: Vec<Digest>`. I can also just compare equality of the roots, since this should imply equality of all the rest. I think this is a slightly weaker test, but don't have strong feelings in this regard.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -689,6 +706,23 @@ pub mod merkle_tree_test {
         assert_eq!(MerkleTreeError::IncorrectNumberOfLeafs, err);
     }
 
+    #[proptest]
+    fn merkle_tree_makers_behave_identically_on_random_input(leafs: Vec<Digest>) {
```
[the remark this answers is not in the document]
jan-ferdinand: The default size range is 0..100. There should be a few Merkle trees among the cases, but the majority will fail.

aszepieniec:
> I'm not fully convinced that the new struct CpuSequential should go into the crate's prelude. Opinions welcome.

Why not? I would treat it the same as `CpuParallel`.

Incidentally, this addition provides an opportunity to get rid of the unnecessary and annoying trait `MerkleTreeMaker` altogether, along with the dysfunctional and poorly named struct `CpuParallel`. What matters is the functions. Any thoughts on that matter?

jan-ferdinand:
I'm happy to extend the scope thusly. Following a rather established naming convention, `MerkleTree::new` could be the newly introduced sequential Merkle tree building, and `MerkleTree::par_new` could be what is currently `MerkleTree::new::<MerkleTreeMaker>` but with the type parameter removed. Objections?

aszepieniec:
:+1:

Sword-Smith:
Can we make it explicit that the sequential constructor is sequential, please?

jan-ferdinand:
That would be `MerkleTree::sequential_new` and `MerkleTree::par_new`, right? This draws attention to any location constructing Merkle trees sequentially, which we assume to be the exception. :+1:

jan-ferdinand:
Do you want to (re-)review the changes?

aszepieniec (approved):

PR #248, opened 2025-01-15
feat : Implement Hasher trait for Tip5

### LINKED ISSUE #207: implement `Hasher` for `Tip5`
To be useful in more contexts, `Tip5` should implement the [`Hasher` trait](https://doc.rust-lang.org/std/hash/trait.Hasher.html).

elielnfinic:
This PR fixes #207

### DIFF
--- a/twenty-first/src/math/tip5.rs
+++ b/twenty-first/src/math/tip5.rs
@@ -1,3 +1,5 @@
+use std::hash::Hasher;
+
 use arbitrary::Arbitrary;
 use get_size2::GetSize;
 use itertools::Itertools;
@@ -673,10 +675,35 @@ impl Sponge for Tip5 {
     }
 }
 
+impl Hasher for Tip5 {
+    fn finish(&self) -> u64 {
+        self.state[0].value()
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        let bfield_elements = bytes.chunks(BFieldElement::BYTES).map(|chunk| {
+            let mut buffer = [0u8; BFieldElement::BYTES];
+            buffer[..chunk.len()].copy_from_slice(chunk);
+            BFieldElement::new(u64::from_le_bytes(buffer))
+        });
+
+        for chunk in bfield_elements.chunks(Tip5::RATE).into_iter() {
+            let mut buffer = [BFieldElement::ZERO; Tip5::RATE];
+            for (buffer_elem, chunk_elem) in buffer.iter_mut().zip(chunk) {
+                *buffer_elem = chunk_elem;
+            }
+            self.absorb(buffer)
+        }
+    }
+}
+
 #[cfg(test)]
 pub(crate) mod tip5_tests {
+    use std::hash::Hash;
     use std::ops::Mul;
 
+    use insta::assert_snapshot;
+    use prop::sample::size_range;
     use proptest::prelude::*;
     use proptest_arbitrary_interop::arb;
     use rand::thread_rng;
@@ -1085,6 +1112,38 @@ pub(crate) mod tip5_tests {
             sponge_cyclomut.state.into_iter().join(","),
             sponge_generated.state.into_iter().join(",")
         );
+    }
+
+    #[test]
+    fn tip5_hasher_trait_test() {
+        let mut hasher = Tip5::init();
+        let data = b"hello world";
+        hasher.write(data);
+        assert_snapshot!(hasher.finish(), @"2267905471610932299");
+    }
+
+    #[proptest]
+    fn tip5_hasher_consumes_small_data(#[filter(!#bytes.is_empty())] bytes: Vec<u8>) {
+        let mut hasher = Tip5::init();
+        bytes.hash(&mut hasher);
+
+        prop_assert_ne!(Tip5::init().finish(), hasher.finish());
+    }
+
+    #[proptest]
+    fn appending_small_data_to_big_data_changes_tip5_hash(
+        #[any(size_range(2_000..8_000).lift())] big_data: Vec<u8>,
+        #[filter(!#small_data.is_empty())] small_data: Vec<u8>,
+    ) {
+        let mut hasher = Tip5::init();
+        big_data.hash(&mut hasher);
+        let big_data_hash = hasher.finish();
+
+        // finish doesn't terminate the hasher; see it's documentation
+        small_data.hash(&mut hasher);
+        let all_data_hash = hasher.finish();
+
+        prop_assert_ne!(big_data_hash, all_data_hash);
     }
 
     #[proptest]


twenty-first/src/math/tip5.rs
```diff
@@ -673,8 +675,38 @@ impl Sponge for Tip5 {
...
+            .collect::<Vec<_>>()
+            .try_into()
+        {
+            self.absorb(chunk);
+        }
+    }
+
+    /// Produce a 64-bit hash value from the sponge's state
+    fn finish(&self) -> u64 {
+        // Extract the first part of the state as the hash value
```
jan-ferdinand, line 702: In my opinion, the comment does not add anything to help understand what's going on.
```suggestion
```

twenty-first/src/math/tip5.rs
```diff
@@ -673,8 +675,38 @@ impl Sponge for Tip5 {
...
+        self.state[0].raw_u64()
```
jan-ferdinand, line 703: Similar to above, I strongly suggest avoiding the `raw` methods.

```suggestion
        self.state[0].value()
```

jan-ferdinand (changes requested):
Thanks for taking this on! Right now, there are some problems with the implementation. They should be relatively easy to resolve, though. :slightly_smiling_face:

jan-ferdinand:
Note also that we're using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for our git commit messages. Before merging this PR, I'd like you to squash the commits and use a message like “feat(Tip5): Implement the `Hasher` trait”. Feel free to change the message if you like. :+1:

twenty-first/src/math/tip5.rs
```diff
@@ -673,8 +675,38 @@ impl Sponge for Tip5 {
...
 
+impl Hasher for Tip5 {
+    fn write(&mut self, bytes: &[u8]) {
+        // Convert input bytes into BFieldElements and absorb them in chunks of `RATE`
+        let mut bfield_elements = bytes
+            .chunks(8) // Assuming each `BFieldElement` corresponds to 8 bytes
+            .map(|chunk| {
+                let mut buffer = [0u8; 8];
+                buffer[..chunk.len()].copy_from_slice(chunk);
+                BFieldElement::from_raw_u64(u64::from_le_bytes(buffer))
```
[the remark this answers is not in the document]
elielnfinic: Okay, thank you for suggesting that.

elielnfinic:
Thanks @jan-ferdinand for your review and suggestions. I hope I was able to resolve your review messages. Please check out this new commit.

twenty-first/src/math/tip5.rs
```diff
@@ -673,10 +675,39 @@ impl Sponge for Tip5 {
...
+            buffer[..chunk.len()].copy_from_slice(chunk);
+            BFieldElement::new(u64::from_le_bytes(buffer))
+        });
+
+        let binding: Vec<BFieldElement> = bfield_elements.collect();
+        let bfield_absorb_chunks = binding.chunks(Tip5::RATE);
+
+        bfield_absorb_chunks.for_each(|absorb_rate_chunk| {
+            let mut asborb_chunk_buffer: [BFieldElement; Tip5::RATE] =
+                [BFieldElement::new(0); Tip5::RATE];
```
jan-ferdinand, line 695: I'm a big fan of type inference and try to let it do whatever it can. How about:

```suggestion
            let mut asborb_chunk_buffer = [BFieldElement::ZERO; Tip5::RATE];
```

The type annotation aside, I personally like to use `BFieldElement::ZERO` over `BFieldElement::new(0)` because it is a constant. On the other hand `BFieldElement::new` is a `const fn` and the compiler might produce the same code in both cases. Both parts of the suggestion – removing the type annotation and using `ZERO` over `new` – are a matter of preference. Feel free to leave things as they are if you prefer that.

twenty-first/src/math/tip5.rs
```diff
@@ -673,10 +675,39 @@ impl Sponge for Tip5 {
...
+
+    fn write(&mut self, bytes: &[u8]) {
+        let bfield_elements = bytes.chunks(BFieldElement::BYTES).map(|chunk| {
+            let mut buffer = [0u8; BFieldElement::BYTES];
+            buffer[..chunk.len()].copy_from_slice(chunk);
+            BFieldElement::new(u64::from_le_bytes(buffer))
+        });
+
+        let binding: Vec<BFieldElement> = bfield_elements.collect();
+        let bfield_absorb_chunks = binding.chunks(Tip5::RATE);
```
jan-ferdinand, line 691: These bindings are not really necessary. Their awkward names suggest that they don't really serve any purpose other than trying to make the types work.

How about starting the second loop in the following way instead?

```rust
for chunk in bfield_elements.chunks(Tip5::RATE).into_iter() {
	// …
}
```

For some reason, the explicit `.into_iter()` seems to be necessary here, even though it usually is not necessary for `for` loops. I guess `itertools::Itertools::IntoChunks` is a little odd?

twenty-first/src/math/tip5.rs
```diff
@@ -673,10 +675,39 @@ impl Sponge for Tip5 {
...
+
+        bfield_absorb_chunks.for_each(|absorb_rate_chunk| {
+            let mut asborb_chunk_buffer: [BFieldElement; Tip5::RATE] =
```
jan-ferdinand, line 694: This variable name feels a litte verbose. I think calling it `buffer`, like you do in the previous loop, works quite well.

twenty-first/src/math/tip5.rs
```diff
@@ -673,10 +675,39 @@ impl Sponge for Tip5 {
...
+                [BFieldElement::new(0); Tip5::RATE];
+            for (i, bfield_elt) in absorb_rate_chunk.iter().enumerate() {
+                asborb_chunk_buffer[i] = *bfield_elt;
+            }
```
jan-ferdinand, line 698: Personally, I like to avoid explicit indexing when I can.

```suggestion
            for (buffer_elt, chunk_elt) in buffer.iter_mut().zip(chunk) {
                *buffer_elt = chunk_elt
            }
```

This one also boils down to personal preference; feel free to ignore this suggestion.

twenty-first/src/math/.tip5.rs.pending-snap
jan-ferdinand, line 1: I don't think this file is needed, since the `insta` snapshot is inline already. Perhaps you committed this by mistake?

jan-ferdinand (commented):
Cool, this is coming together nicely. I've left a few nits inline. Since I have crammed a bunch of comments into a rather tight space, here's what I re-wrote the second loop to on my end. Feel free to pick whetever you like, there's no obligation to take any of it.

```rust
for chunk in bfield_elements.chunks(Tip5::RATE).into_iter() {
    let mut buffer = [BFieldElement::ZERO; Tip5::RATE];
    for (buffer_elem, chunk_elem) in buffer.iter_mut().zip(chunk) {
        *buffer_elem = chunk_elem;
    }
    self.absorb(buffer)
}
```

Lastly, the commit message looks good now. :ok_hand: Can I ask you to also change the commit's description? “Adding test” and “Renaming test” don't really describe what's going on. :wink: You can also leave it empty if you want; I think the description by itself can be sufficient. Your choice. :slightly_smiling_face:

elielnfinic:
I like this suggestion. Thank you.

jan-ferdinand:
This is looking great now! :star:

The only remaining nit I have is the commit's description (the message is great). You can change the description in a number of ways, for example

`git commit --amend -m "feat(Tip5) : Implement the Hasher trait" -m "new description"`

or using an [interactive rebase](https://git-scm.com/docs/git-rebase#_interactive_mode) with the “reword” option.

elielnfinic:
My bad, I believe I had to clean that description the rebase msg file.
Just amended the commit. 

Thank you.

jan-ferdinand:
Thanks for your contribution! :tada:

PR #249, opened 2025-03-26
fetch result after permutation

macro-ss:


jan-ferdinand:
The suggestion changes the meaning of Tip5 in Sponge mode. For example, it doesn't adhere to [the usual definition of a Sponge function](https://en.wikipedia.org/wiki/Sponge_function) anymore; I'd need to do some analysis to know whether the change impacts any security properties. What motivates the change?

Sword-Smith:
Superseeded by #252.

PR #251, opened 2025-04-13
MerkleTree-specific Type Aliases

aszepieniec:
This PR introduces type aliases:
 - `MerkleTreeNodeIndex` for `u64`
 - `MerkleTreeLeafIndex` for `u64`
 - `MerkleTreeHeight` for `u32`.

It also refactors `MerkleTree` to avoid using `usize` where-ever
possible, particularly in public functions. Instead, these types
aliases are used.

`MerkleTree` internally stores a `Vec<Digest>`, which induces two
limitations:
 1. The index type is `usize`, which complicates matters when
    targeting compilation on 32-bit machines (including 32-bit WASM).
 2. This vector can store at most 2^25 nodes, whereas we would like
    functions associated with Merkle trees to work for Merkle trees
    that are far larger (without needing to store all the internal
    nodes explicitly).

The second limitation persists. However, the first limitation is fixed
by this PR.

Also:

 - Public functions `nodes()` and `leafs()` now return iterators
instead of slices. As these functions were public, this change is
breaking.
 - Private function `authentication_structure_node_indices` is
   marked `pub` as it is needed downstream.

Displaces #250.

### DIFF
--- a/twenty-first/src/util_types/merkle_tree.rs
+++ b/twenty-first/src/util_types/merkle_tree.rs
@@ -203,12 +203,15 @@ impl MerkleTree {
         Ok(nodes)
     }
 
-    /// Given a list of leaf indices, return the indices of exactly those nodes that
-    /// are needed to prove (or verify) that the indicated leafs are in the Merkle
-    /// tree.
-    // This function is not defined as a method (taking self as argument) since it's
-    // needed by the verifier, who does not have access to the Merkle tree.
-    fn authentication_structure_node_indices(
+    /// Compute the node indices for an authentication structure.
+    ///
+    /// Given a list of leaf indices, return the indices of exactly those nodes
+    /// that are needed to prove (or verify) that the indicated leafs are in the
+    /// Merkle tree.
+    ///
+    /// Returns an error if any of the leaf indices is bigger than the number of
+    /// leafs.
+    pub fn authentication_structure_node_indices(
         num_leafs: MerkleTreeLeafIndex,
         leaf_indices: &[MerkleTreeLeafIndex],
     ) -> Result<impl ExactSizeIterator<Item = MerkleTreeLeafIndex> + use<>> {
--- a/twenty-first/src/util_types/merkle_tree.rs
+++ b/twenty-first/src/util_types/merkle_tree.rs
@@ -20,28 +20,48 @@ lazy_static! {
             .unwrap_or(DEFAULT_PARALLELIZATION_CUTOFF);
 }
 
-/// Enforces that all compilation targets have a consistent [`MAX_TREE_HEIGHT`].
-/// In particular, if `usize` has more than 32 bits, the maximum height of a
-/// Merkle tree is limited as if only 32 bits were available. If `usize` has
-/// less than 32 bits, compilation will fail.
-///
-/// Using a type other than `usize` could enable a higher maximum height, but
-/// would require a different storage mechanism for the Merkle tree's nodes:
-/// indexing into a `Vec<_>` can only be done with `usize`.
-const MAX_NUM_NODES: usize = 1 << 32;
-const MAX_NUM_LEAFS: usize = MAX_NUM_NODES / 2;
+pub type MerkleTreeNodeIndex = u64;
+pub type MerkleTreeLeafIndex = u64;
+pub type MerkleTreeHeight = u32;
 
-/// The maximum height of a Merkle tree.
-pub const MAX_TREE_HEIGHT: usize = MAX_NUM_LEAFS.ilog2() as usize;
+/// The maximum number of nodes in Merkle trees that functions in this module
+/// support.
+///
+/// This constant enforces that all compilation targets have a consistent
+/// [`MAX_TREE_HEIGHT`] and [`MAX_NUM_LEAFS`].
+///
+/// Note that the struct [`MerkleTree`] can hold only 2^25 nodes. This constant
+/// applies to associated functions that do not take `self`.
+const MAX_NUM_NODES: MerkleTreeNodeIndex =
+    MerkleTreeNodeIndex::MAX ^ (MerkleTreeNodeIndex::MAX >> 1);
+
+/// The maximum number of leafs in Merkle trees that functions in this module
+/// support.
+///
+/// See also: [`MAX_NUM_NODES`], [`MAX_TREE_HEIGHT`].
+const MAX_NUM_LEAFS: MerkleTreeLeafIndex = MAX_NUM_NODES / 2;
+
+/// The maximum height of Merkle trees that functions in this module support.
+///
+/// See also: [`MAX_NUM_NODES`], [`MAX_NUM_LEAFS`].
+pub const MAX_TREE_HEIGHT: MerkleTreeHeight = MAX_NUM_LEAFS.ilog2() as MerkleTreeHeight;
+
+/// The index of the root node.
+pub(crate) const ROOT_INDEX: MerkleTreeNodeIndex = 1;
 
 type Result<T> = result::Result<T, MerkleTreeError>;
 
 /// A [Merkle tree][merkle_tree] is a binary tree of [digests](Digest) that is
 /// used to efficiently prove the inclusion of items in a set. Set inclusion can
-/// be verified through an [inclusion proof](MerkleTreeInclusionProof).
+/// be verified through an [inclusion proof](MerkleTreeInclusionProof). This
+/// struct can hold at most 2^25 digests[^1], limiting the height of the tree to
+/// 2^24. However, the associated functions (*i.e.*, the ones that don't take
+/// `self`) make abstraction of this limitation and work for Merkle trees of up
+/// to 2^63 nodes, 2^62 leafs, or height up to 62.
 ///
 /// The used hash function is [`Tip5`].
 ///
+/// [^1]: https://github.com/Neptune-Crypto/twenty-first/pull/250#issuecomment-2782490889
 /// [merkle_tree]: https://en.wikipedia.org/wiki/Merkle_tree
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct MerkleTree {
@@ -55,7 +75,7 @@ pub struct MerkleTree {
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct MerkleTreeInclusionProof {
     /// The stated height of the Merkle tree this proof is relative to.
-    pub tree_height: usize,
+    pub tree_height: MerkleTreeHeight,
 
     /// The leafs the proof is about, _i.e._, the revealed leafs.
     ///
@@ -63,7 +83,7 @@ pub struct MerkleTreeInclusionProof {
     /// relevant for [`into_authentication_paths`][paths].
     ///
     /// [paths]: MerkleTreeInclusionProof::into_authentication_paths
-    pub indexed_leafs: Vec<(usize, Digest)>,
+    pub indexed_leafs: Vec<(MerkleTreeLeafIndex, Digest)>,
 
     /// The proof's witness: de-duplicated authentication structure for the
     /// leafs this proof is about. See [`authentication_structure`][auth_structure]
@@ -93,16 +113,20 @@ pub struct MerkleTreeInclusionProof {
 /// [auth_structure]: MerkleTree::authentication_structure
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub(crate) struct PartialMerkleTree {
-    tree_height: usize,
-    leaf_indices: Vec<usize>,
-    nodes: HashMap<usize, Digest>,
+    tree_height: MerkleTreeHeight,
+    leaf_indices: Vec<MerkleTreeLeafIndex>,
+    nodes: HashMap<MerkleTreeNodeIndex, Digest>,
 }
 
 impl MerkleTree {
-    /// The index of the root node.
+    /// When iterating over [`Self::nodes`] it pays to have an starting index
+    /// with the same type as the index type used by `Vec`.
     ///
     /// If you need to read the root, try [`root()`](Self::root) instead.
-    pub(crate) const ROOT_INDEX: usize = 1;
+    const ROOT_INDEX: usize = 1_usize;
+
+    const MAX_NUM_NODES: usize = 1_usize << 25;
+    const MAX_NUM_LEAFS: usize = Self::MAX_NUM_NODES / 2;
 
     /// Build a MerkleTree with the given leafs.
     ///
@@ -169,7 +193,7 @@ impl MerkleTree {
         if !num_leafs.is_power_of_two() {
             return Err(MerkleTreeError::IncorrectNumberOfLeafs);
         }
-        if num_leafs > MAX_NUM_LEAFS {
+        if num_leafs > Self::MAX_NUM_LEAFS {
             return Err(MerkleTreeError::TreeTooHigh);
         }
 
@@ -185,9 +209,9 @@ impl MerkleTree {
     // This function is not defined as a method (taking self as argument) since it's
     // needed by the verifier, who does not have access to the Merkle tree.
     fn authentication_structure_node_indices(
-        num_leafs: usize,
-        leaf_indices: &[usize],
-    ) -> Result<impl ExactSizeIterator<Item = usize> + use<>> {
+        num_leafs: MerkleTreeLeafIndex,
+        leaf_indices: &[MerkleTreeLeafIndex],
+    ) -> Result<impl ExactSizeIterator<Item = MerkleTreeLeafIndex> + use<>> {
         // The set of indices of nodes that need to be included in the authentications
         // structure. In principle, every node of every authentication path is needed.
         // The root is never needed. Hence, it is not considered below.
@@ -205,7 +229,7 @@ impl MerkleTree {
             }
 
             let mut node_index = leaf_index + num_leafs;
-            while node_index > Self::ROOT_INDEX {
+            while node_index > ROOT_INDEX {
                 let sibling_index = node_index ^ 1;
                 node_can_be_computed.insert(node_index);
                 node_is_needed.insert(sibling_index);
@@ -251,10 +275,13 @@ impl MerkleTree {
     ///   This is the other part of the de-duplication.
     ///
     /// [verify]: MerkleTreeInclusionProof::verify
-    pub fn authentication_structure(&self, leaf_indices: &[usize]) -> Result<Vec<Digest>> {
+    pub fn authentication_structure(
+        &self,
+        leaf_indices: &[MerkleTreeLeafIndex],
+    ) -> Result<Vec<Digest>> {
         let num_leafs = self.num_leafs();
         let indices = Self::authentication_structure_node_indices(num_leafs, leaf_indices)?;
-        let auth_structure = indices.map(|idx| self.nodes[idx]).collect();
+        let auth_structure = indices.map(|idx| self.node(idx).unwrap()).collect();
         Ok(auth_structure)
     }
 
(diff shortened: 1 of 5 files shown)


jan-ferdinand:
I'm not sure type aliases are the way to go; as aliases, they are not distinct types. If we really want to go that way, I suggest to use the newtype pattern instead.

```rust
pub struct MerkleTreeNodeIndex(u64);

impl MerkleTreeNodeIndex {
    const MAX: u64 = 1 << 63;
}

// similar for these two:
pub struct MerkleTreeLeafIndex(u64);
pub struct MerkleTreeHeight(u8);

// maybe implement std::ops::Deref
```

I understand the primary goal is to enable compilation on 32-bit targets. I think this can be achieved in a simpler way:

```rust
const MAX_NUM_NODES: usize = {
    #[cfg(target_pointer_width = "16")]
    { compile_error!("target pointer width 16 is not supported") }
    #[cfg(target_pointer_width = "32")]
    { 1 << 25 } // comment as to why
    #[cfg(target_pointer_width = "64")]
    { 1 << 32 }
};
```

However, I'm unsure we even want to have this constant. Its documentation states that it “[e]nforces that all compilation targets have a consistent `MAX_TREE_HEIGHT`” – but why is that even important?[^1] Unfortunately, the comment doesn't say. If we really want to have a consistent maximum tree height across compilation targets, and if that maximum tree height should be larger than 24, we need a different approach to our backing storage instead.

Also note that the second limitation you outline, namely that a vector can store at most $2^{25}$ nodes, only applies on 32-bit targets. For 64-bit architectures, the limit is a tree height of 56, or $2^{57}$ nodes.

[^1]: It also enforced this in an unintended way, more akin to `#[cfg(not(target_pointer_width = "64"))] compile_error!("whoups");`.

aszepieniec:
> I'm not sure type aliases are the way to go; as aliases, they are not distinct types. If we really want to go that way, I suggest to use the newtype pattern instead.

I am familiar with the difference between type aliases and newtypes. I do not see the link between type aliases not being distinct types and type aliases being a poor strategy in this particular case or in general. I think you might sneaking in objectives to the discussion that I did not have going in.

Which is fair. The dialectic provoked by different objectives usually exposes superior architectures. But let's be wary of letting the perfect be the enemy of the good.

I realize I was not explicit about my objectives going in, so let me correct that. My goals were:

 - expose `authentication_node_indices` as it is needed downstream
 - enable compilation on 32-bit platforms
 - synchronize node index types across Merkle tree and MMR , particularly by refactoring Merkle tree's `usize`-index into MMR's `u64`-index (I forget the details but this was an irritation in a past episode where I had to use MMR function calls to do Merkle logic because I needed `u64`s)
 - clarify whenever a `u64` is a {Merkle,MMR}-{node,leaf}-index to benefit readability within twenty-first and for dependencies
 - avoid unnecessary downstream refactoring due to dependency-upgrades

Non-goals were:
 - structurally change the code to make it more idiomatic rust
 - fix broken logic
 - make the logic more robust against potential errors

In light of that statement of objectives, I'm not sure why newtypes would be preferable over type aliases.

> However, I'm unsure we even want to have this constant.

Good point! My first thought after reading it stated so succinctly is, "let's kill it". Let me sleep on it to see if other thoughts follow up.

> Also note that the second limitation you outline, namely that a vector can store at most $2^{25}$ nodes, only applies on 32-bit targets. For 64-bit architectures, the limit is a tree height of 56, or $2^{57}$ nodes.

Thanks for the clarification. I'll change the code accordingly.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -179,15 +201,18 @@ impl MerkleTree {
...
+    /// Given a list of leaf indices, return the indices of exactly those nodes
+    /// that are needed to prove (or verify) that the indicated leafs are in the
+    /// Merkle tree.
+    ///
+    /// Returns an error if any of the leaf indices is bigger than the number of
+    /// leafs.
+    pub fn authentication_structure_node_indices(
+        num_leafs: MerkleTreeLeafIndex,
+        leaf_indices: &[MerkleTreeLeafIndex],
+    ) -> Result<impl ExactSizeIterator<Item = MerkleTreeLeafIndex> + use<>> {
```
jan-ferdinand, line 215: This is the reason I'm not a fan of creating aliases for simple types: the function name and the documentation imply `MerkleTreeNodeIndex` for the iterator's item type, yet it is `MerkleTreeLeafIndex`.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -93,16 +111,20 @@ pub struct MerkleTreeInclusionProof {
...
 }
 
 impl MerkleTree {
-    /// The index of the root node.
+    /// When iterating over [`Self::nodes`] it pays to have an starting index
+    /// with the same type as the index type used by `Vec`.
     ///
     /// If you need to read the root, try [`root()`](Self::root) instead.
-    pub(crate) const ROOT_INDEX: usize = 1;
+    const ROOT_INDEX: usize = 1_usize;
```
jan-ferdinand, line 124:
```suggestion
    const ROOT_INDEX: usize = ROOT_INDEX as usize;
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -251,59 +276,75 @@ impl MerkleTree {
...
     }
 
     pub fn root(&self) -> Digest {
         self.nodes[Self::ROOT_INDEX]
     }
 
-    pub fn num_leafs(&self) -> usize {
-        let node_count = self.nodes.len();
+    pub fn num_leafs(&self) -> MerkleTreeLeafIndex {
+        let node_count = MerkleTreeNodeIndex::try_from(self.nodes.len()).expect("usize to u64");
```
jan-ferdinand, line 294: Please use the existing error strings `err::USIZE_TO_U64_ERR` and, when trying to convert a u32 to a usize, `err::U32_TO_USIZE_ERR`.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -251,59 +276,75 @@ impl MerkleTree {
...
         debug_assert!(node_count.is_power_of_two());
         node_count / 2
     }
 
-    pub fn height(&self) -> usize {
+    pub fn height(&self) -> MerkleTreeHeight {
         let leaf_count = self.num_leafs();
         debug_assert!(leaf_count.is_power_of_two());
-        leaf_count.ilog2() as usize
+        MerkleTreeHeight::try_from(leaf_count.ilog2()).expect("log of num leafs should fit in u32")
```
jan-ferdinand, line 302: `MerkleTreeHeight` is a type alias for u32, `.ilog2()` produces a u32.

```suggestion
        leaf_count.ilog2()
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -251,59 +276,75 @@ impl MerkleTree {
...
     }
 
     /// All nodes of the Merkle tree.
-    pub fn nodes(&self) -> &[Digest] {
-        &self.nodes
+    pub fn nodes(&self) -> impl Iterator<Item = &Digest> {
+        self.nodes.iter()
```
jan-ferdinand, line 307: How about skipping the first element? The dummy node there seems like an implementation detail to me that we shouldn't expose.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -251,59 +276,75 @@ impl MerkleTree {
...
     }
 
     /// The node at the given node index, if it exists.
-    pub fn node(&self, index: usize) -> Option<Digest> {
-        self.nodes.get(index).copied()
+    pub fn node(&self, index: MerkleTreeNodeIndex) -> Option<Digest> {
+        usize::try_from(index)
+            .ok()
+            .and_then(|idx| self.nodes.get(idx).copied())
```
jan-ferdinand, line 314: Maybe add a dev comment why this works?

```rust
// If `MerkleTreeNodeIndex` aka u64 cannot be converted to usize, that means
// (1) the current architecture has a pointer width smaller than 64 bits and
// (2) the current index is larger than the pointer width. Therefore, Merkle
// trees with the number of nodes implied by the requested index cannot even
// constructed on the current architecture.
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -251,59 +276,75 @@ impl MerkleTree {
...
 
     /// All leafs of the Merkle tree.
-    pub fn leafs(&self) -> &[Digest] {
-        let first_leaf = self.nodes.len() / 2;
-        &self.nodes[first_leaf..]
+    pub fn leafs(&self) -> impl Iterator<Item = &Digest> {
+        self.nodes.iter().skip(
+            (self.num_nodes() / 2)
+                .try_into()
+                .expect("MerkleTreeNodeIndex to usize conversion error"),
```
jan-ferdinand, line 322:
```suggestion
        // This conversion can only fail if the number of leafs is larger than
        // usize::MAX. This implies that the number of nodes is larger than
        // usize::MAX. Since the nodes are stored in a Vec, the number of nodes
        // can never exceed usize::MAX.
        // This proof by contradiction shows that unwrapping is fine.
        let num_leafs = usize::try_from(self.num_leafs()).unwrap();

        self.nodes.iter().skip(num_leafs)
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -251,59 +276,75 @@ impl MerkleTree {
...
         let num_leafs = self.num_leafs();
         let invalid_index = MerkleTreeError::LeafIndexInvalid { num_leafs };
         let maybe_indexed_leaf = |&i| self.leaf(i).ok_or(invalid_index).map(|leaf| (i, leaf));
 
         indices.iter().map(maybe_indexed_leaf).collect()
     }
 
+    fn num_nodes(&self) -> MerkleTreeNodeIndex {
+        MerkleTreeNodeIndex::try_from(self.nodes.len())
+            .expect("`MerkleTreeNodeIndex` should be large enough to hold any usize")
```
jan-ferdinand, line 345: Maybe use the existing `err::USIZE_TO_U64_ERR` string? Here, I'm less sure.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -608,8 +656,15 @@ pub mod merkle_tree_test {
...
-        #[strategy(vec(0..#tree.num_leafs(), 0..#tree.num_leafs()))]
-        pub selected_indices: Vec<usize>,
+        #[
+            strategy(
+                vec(
+                    (0 as MerkleTreeLeafIndex)..#tree.num_leafs(),
+                    0..(#tree.num_leafs() as usize)
+                )
+            )
+        ]
```
jan-ferdinand, line 666: There's no need for the first cast, since `#tree.num_leafs()` is already of type `MerkleTreeLeafIndex`.

```suggestion
        #[strategy(vec(0..#tree.num_leafs(), 0..(#tree.num_leafs() as usize)))]
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -179,15 +201,18 @@ impl MerkleTree {
         Ok(nodes)
```
jan-ferdinand, line 201: After the `MAX_*` constants are gone, this could become something like the following:

```rust
let num_nodes = 2 * num_leafs;
let mut nodes = Vec::new();
nodes
    .try_reserve_exact(num_nodes)
    .map_err(|_| MerkleTreeError::TreeTooHigh)?;
nodes.resize(num_nodes, Digest::default());
nodes[num_leafs..].copy_from_slice(leafs);

Ok(nodes)
```

Overcommitting memory allocators might throw wrenches in gears. ⚙️🔧⚙️

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -394,19 +435,19 @@ impl MerkleTreeInclusionProof {
...
             .get(&index)
             .copied()
             .ok_or(MerkleTreeError::MissingNodeIndex(index))
     }
 
-    fn num_leafs(&self) -> Result<usize> {
+    fn num_leafs(&self) -> Result<MerkleTreeLeafIndex> {
         if self.tree_height > MAX_TREE_HEIGHT {
             return Err(MerkleTreeError::TreeTooHigh);
         }
```
jan-ferdinand, line 453: After the `MAX_*` constants are gone, this could become something like:

```rust
1_u64
    .checked_shl(self.tree_height)
    .ok_or(MerkleTreeError::TreeTooHigh)
```

jan-ferdinand (commented):
I think removing the various `MAX_*` constants is the way to go.

jan-ferdinand:
We should probably add a CI workflow (or modify an existing one) to check that compilation on 32-bit architectures works. The following is a starting point:

```sh
rustup target add i686-unknown-linux-gnu
cargo clippy --target i686-unknown-linux-gnu
```

I'm not familiar with all the pitfalls of cross compilation, but this seems like a starting point.

twenty-first/src/util_types/merkle_tree.rs
aszepieniec: Good catch! This should absolutely be `MerkleTreeNodeIndex`.

I understand the drawback of not having the compiler's help to identify inconsistencies. It leads to wrong aliases slipping through the cracks, like here. But if that's a compelling argument against using type aliases, then it's also a compelling argument against comments in the code describing what's going on. Do you have a similar policy on comments (not including doctests or docstrings)?

twenty-first/src/util_types/merkle_tree.rs
aszepieniec: Doesn't proptest complain about ranges whose type is only determined after the type inference has run? I'm happy to give this commit a try but if it ends up being reverted you know why ;-)

twenty-first/src/util_types/merkle_tree.rs
aszepieniec: Consider this example:

```rust
if merkle_tree.nodes()[1] == expected_root {
    rejoice!("hooray!");
}
```

Which is obviously a complete violation of interface boundaries but it wouldn't surprise me if there were lines like this all over the code base. Bite the bullet?

twenty-first/src/util_types/merkle_tree.rs
aszepieniec: I don't see why not. More specifically, I don't see why we should treat the two cases where this conversion happens differently.

twenty-first/src/util_types/merkle_tree.rs
aszepieniec: We also have a function called `num_nodes()` which should probably agree with the number of elements that the iterator can yield ...

aszepieniec: d04e8aa544ba0bd65dca201afb861848cbc5f03b bites the bullet thusly:
 - kill `nodes()` :dagger: 
 - kill `num_nodes()` :dagger:

jan-ferdinand: `tasm-lib` uses `.nodes()` in [one place](https://github.com/search?q=repo%3ATritonVM%2Ftasm-lib%20.nodes()&type=code), as far as I can tell. This one place doesn't seem critical to performance, meaning it can easily switch to iterating the nodes' indices and requesting the corresponding nodes one-by-one. :+1:

twenty-first/src/util_types/merkle_tree.rs
jan-ferdinand: To a certain degree. If the function's signature (including name) fully explains what the function does, then I don't add the noise that a comment would now be. However, since the comment has to follow fewer rules than the name of the function, it's easy to add information not captured in the signature, which makes adding a comment worthwhile often enough.

All that said, I might misunderstand what you're getting at, since you are excluding docstrings from consideration. The type aliases are public, as should the description of a function's “what” be, meaning it should go in the docstring.

.github/workflows/main.yml
```diff
@@ -41,6 +41,9 @@ jobs:
         env:
           RUSTDOCFLAGS: -D warnings
 
+      - name: Install targets
+        run: rustup target add i686-unknown-linux-gnu wasm32-unknown-unknown
+
```
jan-ferdinand, line 46: Even though the step below has argument `--all-targets`, that means a [different kind of target](https://doc.rust-lang.org/cargo/commands/cargo-check.html#target-selection)[^clippydoc], like `lib` or `bin` or `example`. To actually use the rust toolchains for the 32-bit architectures, this step could instead be:

```suggestion
      - name: Build for 32-bit architectures
        run: |
          rustup target add i686-unknown-linux-gnu wasm32-unknown-unknown
          cargo build --target i686-unknown-linux-gnu
          cargo build --target wasm32-unknown-unknown
```

I'm not sure it's a good idea to include this step is the `main.yml` workflow, as it is run on all three major operating systems, creating duplicate work and slowing down CI. It might be better to add a new CI workflow instead. Alternatively, this step can execute conditional to the current `matrix.os`.

[^clippydoc]: `cargo clippy --help` points to `cargo check --help` which is equivalent to the linked page.

aszepieniec: Thanks for the correction. How do you recommend to proceed?

jan-ferdinand: The first thing that comes to mind is to create a new CI workflow specifically for building the crate on a 32-bit architecture. This keeps the individual workflows _relatively_ lean, performs the new step only once instead of once per element in the strategy's matrix, at the expense of having a new workflow. Such a workflow could look like this:

```yaml
name: 32bit

on:
  push:
    branches:
      - master
  pull_request:
    branches:
      - master

jobs:
  build:
    name: Build for 32-bit targets
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v4

      - name: Install stable toolchain
        uses: dtolnay/rust-toolchain@stable

      - name: Build for 32-bit architectures
        run: |
          rustup target add i686-unknown-linux-gnu wasm32-unknown-unknown
          cargo build --target i686-unknown-linux-gnu
          cargo build --target wasm32-unknown-unknown

```

twenty-first/src/util_types/merkle_tree.rs
aszepieniec: Type aliases can have docstrings, and these particular docstrings can be used to document index conventions. This solution strikes me as better than the alternative, which is to document the relevant bits of the index conventions in the functions' docstrings. This alternative leads to a scattered convention documentation.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -351,7 +340,7 @@ impl MerkleTree {
         // usize::MAX. Since the nodes are stored in a Vec, the number of nodes
         // can never exceed usize::MAX.
         // This proof by contradiction shows that unwrapping is fine.
-        let num_leafs = usize::try_from(self.num_leafs()).unwrap();
+        let num_leafs = self.num_leafs();
 
         self.nodes.iter().skip(num_leafs)
```
jan-ferdinand, line 345: The comment arguing that `.unwrap()` is fine is no longer necessary.

```suggestion
        self.nodes.iter().skip(self.num_leafs())
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -674,7 +664,7 @@ pub mod merkle_tree_test {
         ) -> HashMap<MerkleTreeNodeIndex, Digest> {
             node_indices
                 .iter()
-                .map(|&i| (i, BFieldElement::new(i)))
+                .map(|&i| (i, BFieldElement::new(u64::try_from(i).unwrap())))
```
jan-ferdinand, line 667: How about the `bfe!` macro?

```suggestion
                .map(|&i| (i, bfe!(i))))
```

twenty-first/src/util_types/merkle_tree.rs
aszepieniec: Fixed in 4d7fc1718cdab39106900993d1155d15b6b49a21.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -842,7 +911,8 @@ pub mod merkle_tree_test {
     #[proptest(cases = 30)]
     fn requesting_inclusion_proof_for_nonexistent_leaf_fails_with_expected_error(
         #[strategy(arb())] tree: MerkleTree,
-        #[filter(#leaf_indices.iter().any(|&i| i > #tree.num_leafs()))] leaf_indices: Vec<usize>,
+        #[filter(#leaf_indices.iter().any(|&i| i > (#tree.num_leafs() as MerkleTreeLeafIndex)))]
```
jan-ferdinand, line 914: `.num_leafs()` already returns type `MerkleTreeLeafIndex`.

```suggestion
        #[filter(#leaf_indices.iter().any(|&i| i > #tree.num_leafs()))]
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -901,11 +971,11 @@ pub mod merkle_tree_test {
     #[proptest(cases = 10)]
     fn each_leaf_can_be_verified_individually(test_tree: MerkleTreeToTest) {
         let tree = test_tree.tree;
-        for (leaf_index, &leaf) in tree.leafs().iter().enumerate() {
+        for (leaf_index, &leaf) in tree.leafs().enumerate() {
             let authentication_path = tree.authentication_structure(&[leaf_index]).unwrap();
             let proof = MerkleTreeInclusionProof {
                 tree_height: tree.height(),
-                indexed_leafs: [(leaf_index, leaf)].into(),
+                indexed_leafs: [(leaf_index as MerkleTreeLeafIndex, leaf)].into(),
```
jan-ferdinand, line 978: Depends on how hard you want to use `MerkleTreeLeafIndex` or accept that it is an alias for `usize`.

```suggestion
                indexed_leafs: [(leaf_index, leaf)].into(),
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -950,7 +1020,7 @@ pub mod merkle_tree_test {
         //
         //  0      2   <-- opened_leaf_indices
 
-        let node_indices = [3, 8, 9, 10, 11];
+        let node_indices = [3 as MerkleTreeNodeIndex, 8, 9, 10, 11];
```
jan-ferdinand, line 1023: Type inference can figure this out by itself.

```suggestion
        let node_indices = [3, 8, 9, 10, 11];
```

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -394,23 +449,22 @@ impl MerkleTreeInclusionProof {
...
             .copied()
             .ok_or(MerkleTreeError::MissingNodeIndex(index))
     }
 
-    fn num_leafs(&self) -> Result<usize> {
-        if self.tree_height > MAX_TREE_HEIGHT {
-            return Err(MerkleTreeError::TreeTooHigh);
-        }
-        Ok(1 << self.tree_height)
+    fn num_leafs(&self) -> Result<MerkleTreeLeafIndex> {
```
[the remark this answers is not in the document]
jan-ferdinand: To me, `num_x` implies the result be a cardinal number, but `x_index` sounds like an ordinal number. `usize` doesn't claim to be either, and so to me it feels like it can be used for both without triggering this feeling of dissonance.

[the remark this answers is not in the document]
aszepieniec: How is "index" ordinal? O_o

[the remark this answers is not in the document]
jan-ferdinand: It might be more a language thing than a math thing, but since this PR is primarily about names, that feels in scope. “In linguistics, ordinal numerals […] are words representing position or rank in a sequential order.” ([source](https://en.wikipedia.org/wiki/Ordinal_numeral))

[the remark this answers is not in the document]
jan-ferdinand: Querying types like [`HashMap`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.len) or [`str`](https://doc.rust-lang.org/std/primitive.str.html#method.len) for their length results in a `usize`, even though they (generally) cannot be indexed with that type. To me, this indicates that (1) the first option is consistent with prominent rust uses like the standard library and (2) changing the index type does not imply that the return type for `num_*()` functions would change.

In any case, options 1, 3, and 4 all seem (to varying degrees) fine to me.

PR #257, opened 2025-05-28
feat: Compute Merkle root with minimal RAM usage

jan-ferdinand:
- [x] compute Merkle root using as little RAM as possible
- [x] compute Merkle root in parallel using little RAM
- [x] compute authentication structure using as little RAM as possible
- [x] compute authentication structure in parallel using little RAM

### DIFF
--- a/twenty-first/src/util_types/merkle_tree.rs
+++ b/twenty-first/src/util_types/merkle_tree.rs
@@ -335,6 +335,10 @@ impl MerkleTree {
         num_leafs: MerkleTreeLeafIndex,
         leaf_indices: &[MerkleTreeLeafIndex],
     ) -> Result<impl ExactSizeIterator<Item = MerkleTreeNodeIndex> + use<>> {
+        if !num_leafs.is_power_of_two() {
+            return Err(MerkleTreeError::IncorrectNumberOfLeafs);
+        }
+
         // The set of indices of nodes that need to be included in the authentications
         // structure. In principle, every node of every authentication path is needed.
         // The root is never needed. Hence, it is not considered below.


Sword-Smith:
I wonder if you've just re-implemented MMR-accumulator. Maybe that code could be reused?

twenty-first/benches/merkle_tree.rs
```diff
@@ -8,20 +8,37 @@ use twenty_first::math::digest::Digest;
...
 use twenty_first::util_types::merkle_tree::MerkleTree;
 
-fn merkle_tree(c: &mut Criterion) {
-    let mut group = c.benchmark_group("merkle_tree");
+criterion_main!(benches);
+criterion_group!(
+    name = benches;
+    config = Criterion::default()
+        .sample_size(10)
+        .measurement_time(Duration::from_secs(50));
```
Sword-Smith, line 16: I think this line is not needed, and that the benchmark runs faster if removed.

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -189,6 +198,94 @@ impl MerkleTree {
...
+        Ok(root)
+    }
+
+    /// Compute the Merkle root from the given leafs in parallel without
+    /// recording any internal nodes.
+    ///
+    /// This is equivalent to
+    /// [`MerkleTree::par_new`]`(leafs).map(|t| t.`[`root()`][root]`)`
+    /// but requires considerably less RAM without impacting runtime performance
+    /// (it neither improves nor worsens). See [`Self::sequential_frugal_root`]
```
Sword-Smith, line 237: Not accurate. This algorithm is faster than `Merkletree::par_new` :)

twenty-first/src/util_types/merkle_tree.rs
```diff
@@ -189,6 +198,94 @@ impl MerkleTree {
...
+            // at latest once num_threads equals 1.
+            while num_threads > leafs.len() / 2 {
+                num_threads /= 2;
+            }
+
+            let chunk_size = leafs.len() / num_threads;
+            let next_layer = (0..num_threads)
+                .into_par_iter()
+                .map(|i| Self::sequential_frugal_root(&leafs[i * chunk_size..(i + 1) * chunk_size]))
+                .collect::<Result<_>>()?;
```
Sword-Smith, line 281: Very cool!

jan-ferdinand:
> The sequential recalculation of the auth structure looks really slow.

I think that's due to the size of the Merkle tree that was used in the benchmark. Previously, it had height 22. I have changed it to have height 20 at most, reflecting the `merkle_tree.rs` benchmark.

jan-ferdinand:
> Nit pick: Benchmarks for merkle_tree_authenticate could also have their sizes reflected in the benchmark names -- like the merkle_tree benchmarks have.

Yeah, that's a good improvement. Adressed in 5a19d45dba089e676c57652e46719eb19238651f.

PR #259, opened 2025-06-06
perf: Speed up parallel Merkle tree building

jan-ferdinand:
After having observed the performance difference between RAM-hungry and RAM-frugal root computation in #257, I got perf hungry. This PR brings some of that performance boost to RAM-hungry Merkle tree building.

The commits are self-contained and it might make sense to review them in isolation.

02720bef: Where previously, parallelization happened only within a given tree layer, now multiple subtrees are computed in parallel.

9e96e848: The underlying `nodes` vector is initialized in parallel. This is a less clear-cut performance boost, as it is a slowdown for small trees (height roughly less than 18) but a definitive win for trees larger than that. (The exact numbers depend on the concrete hardware.) This commit introduces some `unsafe` code and should be reviewed with the appropriate diligence.

e9fae6aa: A change not directly related to performance. Where previously, the parallelization cutoff could only be set via an environment variable, it can now also be set programatically, using the new public module `config`. The environment variable has precedence. The change allows testing different values for the cutoff, which was previously impossible. The commit also introduces tests that use that new functionality.

### DIFF
--- a/twenty-first/src/config.rs
+++ b/twenty-first/src/config.rs
@@ -1,0 +1,77 @@
+//! This module contains various configuration options. In general, the
+//! configuration options impact performance only. The default configuration is
+//! sane and should provide good performance for most users.
+//!
+//! Most configuration options can also be set via environment variables.
+//! Generally, the environment variables take precedence over the options set
+//! in this module.
+
+use std::cell::RefCell;
+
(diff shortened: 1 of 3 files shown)


Sword-Smith (approved):
Nice speedup!!

PR #268, opened 2025-06-24
Vectorize Tip5

jan-ferdinand:
On machines that support it, and using conditional compilation, Tip5 can be vectorized using AVX-512, which results in significant performance improvements. Credit goes to the team at https://kompilers.com/

## How To

To build the new goodies, use:

`RUSTFLAGS="-C target-feature=+avx512ifma,+avx512f,+avx512bw,+avx512vbmi" cargo build`

If you are compiling on a machine that has all the necessary features, you can use the following instead:

`RUSTFLAGS="-C target-cpu=native" cargo build`

## Performance

You can expect the following raw performance gains:

```
hash_10                 time:   [336.29 ns 336.92 ns 337.72 ns]
                        change: [−31.354% −31.251% −31.139%] (p = 0.00 < 0.05)
                        Performance has improved.

hash_varlen/len/16384   time:   [526.63 µs 527.02 µs 527.51 µs]
                        change: [−32.321% −32.151% −32.017%] (p = 0.00 < 0.05)
                        Performance has improved.

hash_parallel/len/65536 time:   [2.0350 ms 2.0502 ms 2.0656 ms]
                        change: [−10.764% −8.6165% −5.8573%] (p = 0.00 < 0.05)
                        Performance has improved.
```

In addition, there are now specialized functions for SIMD hashing. Since those functions didn't exist before, they cannot be compared quite as easily. For example, we have `hash_varlen_many`:

```
hash_varlen_many/(len, N)/(16384, 2)
                        time:   [783.82 µs 787.46 µs 791.45 µs]
```

This is a ~25% speedup over the already sped-up `hash_varlen` (794 μs vs 1054 μs = 2·527 μs). Note that using these new functions requires changes in the calling code, whereas no changes are needed to get the performance benefits of the functions that existed already.

All benchmarks performed on the machine known as megingjord.

## Gotchas

Note that rust is somewhat conservative when it comes to enabling CPU-architecture specific features. In particular, even if you compile this code on a CPU that has all the necessary AVX features, rust opts to not enable them in anticipation of the binary being distributed. Thus, you must pass `RUSTFLAGS="-C target-cpu=native"` to get access to the vectorized version of Tip5. In the future, we can use [dynamic dispatch](https://doc.rust-lang.org/std/arch/#dynamic-cpu-feature-detection) instead, but this PR doesn't include that.

Note also that github's CI runners do not have AVX-512 capabilities (or at least, they are not exposing them in a reliable way). This means that the new code is _not_ tested in CI. If you know a way to get that working, I'd be happy to learn of it.

### DIFF
--- a/twenty-first/src/tip5/mod.rs
+++ b/twenty-first/src/tip5/mod.rs
@@ -901,7 +901,6 @@ pub(crate) mod tests {
     use prop::sample::size_range;
     use proptest::prelude::*;
     use proptest_arbitrary_interop::arb;
-    use rand::Rng;
     use rand::RngCore;
     use rayon::prelude::IntoParallelIterator;
     use rayon::prelude::ParallelIterator;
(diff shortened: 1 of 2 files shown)


Sword-Smith:
@Holindauer We got some AVX512 code here with a discrepancy. @jan-ferdinand is looking at it but maybe you have some insight?

PR #270, opened 2025-09-30
Add & change formatting options to `BFieldElement`

jan-ferdinand:
- change `Debug` implementation to show canonical representation, not Montgomery representation
- add implementations for `LowerHex` and `UpperHex` formatting traits


Sword-Smith (approved):
Looks good to me. I've never had to use a Montgomery representation for anything when reading a debug or display output, so I don't think it'll be missed.

PR #274, opened 2025-12-03
Make sure that Tip5 rounds do not produce degenerate elements

### LINKED ISSUE #273: Missing modulo prime step in the mds layer
In `mds_generated`, there is a missing modulo prime step. 
See the following reproduction where the naive and real implementations produce a different result.

```rust
#[test]
fn test_mds_generated_corresponds_to_mds_matrix_mul() {
    let state = [
        15567475466831782388,
        14121273707351390936,
        522093363340240805,
        11776204468718162041,
        5418461429836749777,
        13219597586193145899,
        10838844705436671494,
        5101097820672629560,
        10054772990141757372,
        10490664012066605149,
        4779488651980666516,
        6723093447186088526,
        3165195137037532367,
        2825033159687270366,
        13665968981428010166,
        3894747349555678880,
    ].map(BFieldElement::from_raw_u64);

    let mut naive = NaiveTip5 { state };
    let mut real = Tip5 { state };

    assert_eq!(naive, real);

    naive.mds_matrix_mul();
    real.mds_generated();

    assert_eq!(naive, real);
}
```

jan-ferdinand:
I’m scared of changing the parts of Tip5 I don’t fully understand. The highly optimized version of MDS matrix multiplication is one such part. Even though #273 identifies a bug in the implementation, I rather add a bunch of comments explaining the current status as well as various tests to ensure the identified invariants than to touch the code. Since the buggy method is not public, I consider this PR to fix #273.

This PR also adds an inverse of Tip5, which I used for debugging. It’s not needed in production and only lives under the `test` flag. It might assist in future Tip5 debugging sessions, of which I hope there will be few.

### DIFF
--- a/twenty-first/src/tip5/mod.rs
+++ b/twenty-first/src/tip5/mod.rs
@@ -41,6 +41,8 @@ pub mod digest;
 ))]
 mod avx512;
 #[cfg(test)]
+mod inverse;
+#[cfg(test)]
 mod naive;
 
 /// The lookup table with a high algebraic degree used in the TIP-5 permutation. To verify its
--- a/twenty-first/src/tip5/mod.rs
+++ b/twenty-first/src/tip5/mod.rs
@@ -213,22 +213,42 @@ impl Tip5 {
         for i in 0..STATE_SIZE {
             let b = self.state[i].raw_u64();
             hi[i] = b >> 32;
-            lo[i] = b & 0xffffffffu64;
+            lo[i] = b & 0xffff_ffff;
         }
 
         lo = Self::generated_function(lo);
         hi = Self::generated_function(hi);
 
+        // In isolation, the following reduction modulo BFieldElement::P is
+        // buggy. Concretely, it can produce degenerate Montgomery
+        // representations, that is, `state[r].0` can be greater than or equal
+        // to BFieldElement::P. While there are many inputs for which this can
+        // happen, the easiest with which to trace the behavior manually is:
+        // lo[r] = 0x10;
+        // hi[r] = 0xf_ffff_fff0;
+        //
+        // These starting values lead to `s_hi = 0`, `s_lo = BFieldElement::P`.
+        // Since `s_hi` is 0, the `overflowing_add` does nothing, and the
+        // degenerate representation from `s_lo` is directly transferred to the
+        // state element.
+        //
+        // All that said, due to the specific context this method is always (!)
+        // used in, the bug does not propagate. In particular, this method is
+        // followed up with round-constant addition. Due to a quirk in the
+        // implementation of `BFieldElement::Add` and a property of the round
+        // constants, any degenerate representations are corrected.
+        //
+        // Below, you can find tests for the specific properties claimed. The
+        // doc-string of those tests mention the name of this method.
         for r in 0..STATE_SIZE {
             let s = (lo[r] >> 4) as u128 + ((hi[r] as u128) << 28);
 
             let s_hi = (s >> 64) as u64;
             let s_lo = s as u64;
 
-            let (res, over) = s_lo.overflowing_add(s_hi * 0xffffffffu64);
+            let (res, over) = s_lo.overflowing_add(s_hi * 0xffff_ffff);
 
-            self.state[r] =
-                BFieldElement::from_raw_u64(if over { res + 0xffffffffu64 } else { res });
+            self.state[r] = BFieldElement::from_raw_u64(if over { res + 0xffff_ffff } else { res });
         }
     }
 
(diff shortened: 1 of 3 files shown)


twenty-first/src/tip5/mod.rs
```diff
@@ -1067,6 +1089,50 @@ pub(crate) mod tests {
         assert_eq!(Tip5::offset_fermat_cube_map(255), 255);
     }
 
+    /// Ensure that the claims made in [`Tip5::mds_generated`] are true.
+    ///
+    /// In particular, `BFieldElement::Add` internally uses the equality
+    /// `a + b = a - (p - b)`. If `a` is a degenerate representation (i.e., the
+    /// Montgomery representation is larger than or equal to
+    /// [`BFieldElement::P`]), then a small enough `b` makes the result of the
```
aszepieniec: The parenthesis does not make sense to me. If the number if larger than or equal to $p$, then it is not in Montgomery representation.

twenty-first/src/tip5/mod.rs
```diff
@@ -1067,6 +1089,50 @@ pub(crate) mod tests {
...
+    /// addition non-degenerate by removing the “surplus” from `a`. In
+    /// particular, this correction will happen if the following inequality
+    /// holds:
+    ///              p - b > u64::MAX - p
+    ///   ↔          p - b > u32::MAX
+    ///   ↔   p - u32::MAX > b
```
aszepieniec: I get a much tighter bound: `p - b > u32::MAX - 1`. Because expanding `u64::MAX - p = 2^64 - 1 - 2^64 + 2^32 - 1 = 2^32 - 1 - 1`.

twenty-first/src/tip5/mod.rs
```diff
@@ -1067,6 +1089,50 @@ pub(crate) mod tests {
...
+    ///
+    /// While it’s not particularly beautiful to depend on such implementation
+    /// details, Ferdinand is too scared to change the implementation of Tip5.
+    /// If you change the behavior of `BFieldElement::Add`, please make sure
+    /// that [`Tip5`] is not breaking.
```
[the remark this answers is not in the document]
aszepieniec: I mean, they will obviously discover as a result of this test, but getting a Tip5 test vector test to fail drives the message home harder.

twenty-first/src/tip5/mod.rs
```diff
@@ -1067,6 +1089,50 @@ pub(crate) mod tests {
...
+    ///
+    /// The test [`round_constants_correct_degenerate_lhs_when_adding`] makes
+    /// sure that all round constants have the required property.
+    ///
+    /// See also: https://www.hyrumslaw.com/. Sorry about that.
+    #[proptest]
+    fn adding_degenerate_lhs_and_small_enough_rhs_makes_lhs_non_degenerate(
+        #[strategy(BFieldElement::P..)] degenerate_left: u64,
+        #[strategy(0..BFieldElement::P - u64::from(u32::MAX))] raw_right: u64,
```
aszepieniec: I don't see how this strategy is correct.

I would naïvely expect the variables that take the values of `a` and `b` in the above text to be called `a` and `b`. But now there is a renaming of variables and algebraic manipulations going on, and that's too much strain for my little brain.

twenty-first/src/tip5/naive.rs
```diff
@@ -51,8 +51,9 @@ impl NaiveTip5 {
         for (row_idx, new_elem) in new_state.iter_mut().enumerate() {
             for col_idx in 0..STATE_SIZE {
                 // See <https://en.wikipedia.org/wiki/Circulant_matrix>
-                let first_col_index = (col_idx * (STATE_SIZE - 1) + row_idx) % STATE_SIZE;
-                let matrix_element = BFieldElement::from(MDS_MATRIX_FIRST_COLUMN[first_col_index]);
+                // The initial summand `STATE_SIZE` only prevents overflows.
+                let mds_matrix_idx = (STATE_SIZE + row_idx - col_idx) % STATE_SIZE;
```
aszepieniec: Underflows? Either way, big improvement :clapper:

aszepieniec (changes requested):

twenty-first/src/tip5/mod.rs
jan-ferdinand: Maybe it becomes more clear when I remove the words “the Montgomery representation”?

“If `a` is a degenerate representation (i.e., is larger than or equal to [`BFieldElement::P`]), then […]”

twenty-first/src/tip5/mod.rs
jan-ferdinand: I.e., `p - u32::MAX + 1 > b`? That’s indeed much tighter. 😁 Yes, seems like I’m off by one.

aszepieniec (approved):

PR #276, opened 2025-12-04
Deprecate `XFieldElement::{in, de}crement`

jan-ferdinand:
While `BFieldElement::{in, de}crement` are used in downstream crates, the corresponding `XFieldElement` methods are not. With the additional argument, they are even prone to panics.


Sword-Smith (approved):
If not used downstream, let's kill! LGTM.

